PR #122, opened 2014-11-04
Return 404 status code if custom 404 error page exists

OliverJAsh:


OliverJAsh:
Did we ever do a clean up of this? /cc @jesusabdullah

jesusabdullah:
No, I never got around to digging deeper.

PR #129, opened 2015-02-04
Include .npmignore file

### LINKED ISSUE #128: Special characters in test filenames causing issues with CI server
Jenkins has a problem clearing out the workspace if there are files with special characters. This tests triggers that problem: 檔案.html

losttime:
If no `.npmignore` file is included, `.gitignore` is used by default. Thus, everything in `.gitignore` would typically end up in the `.npmignore` file.

The `.npmignore` file I've added uses the `.gitignore` as a base, and adds the `test` directory to resolve #128. Further, I've added a note to the `.gitignore` to help ensure future modifications there get added to the `.npmignore` too.

With this change, those that `npm install` the package won't get the tests, but they likely don't need them. If they're going to be running tests against this module, they'll likely be cloning the repo in the first place.

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -15,3 +15,6 @@ results
 
 node_modules
 npm-debug.log
+
+# NPM-specific (in git, but not downloaded with `npm install`)
+test


losttime:
What do you think @jesusabdullah ?

jfhbrook:
Let me sit on this one for a while. I feel your pain with #128 but I also think including test files with the package is a good idea. I can launch into the philosophical discussion if you'd like, but it's past 1:00am for me.

I'll probably tackle PRs this weekend, it looks like I have a few.

jfhbrook:
Merged in https://github.com/jfhbrook/node-ecstatic/commit/ebac68d28ddd650daa374a83e1f0ed13722283aa

PR #130, opened 2015-02-10
Add option to disable 'Server' header

SomeoneWeird:
As the title says, this PR adds the ability to disable the 'Server' header for all responses that ecstatic handles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,8 @@ var opts = {
              humanReadable : true,
              si            : false,
              defaultExt    : 'html',
-             gzip          : false
+             gzip          : false,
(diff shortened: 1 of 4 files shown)


jfhbrook:
Can you elaborate on why this is necessary? I'm open to it, I just don't want to needlessly increase the size of an already significant API surface area.

SomeoneWeird:
Well, it's a slight security concern, as people poking around could easily see that header and come to the conclusion that the server is running node (granted, there are other ways too, but a header is by far the obvious). It also adds a (very) insignificant size to the payload.

PR #133, opened 2015-02-27
Bugfix: `handleErrors` is ignored when using `showDir`

jantimon:
I would like to use the *proxy-middleware* if the request cannot be resolved locally.
However right now `handleErrors = false` is ignored if `showDir` is set to `true` and the directory does not exists and displays a error page instead.

### DIFF
--- a/lib/ecstatic/showdir.js
+++ b/lib/ecstatic/showdir.js
@@ -12,6 +12,7 @@ module.exports = function (opts, stat) {
       root = path.resolve(opts.root),
       baseDir = opts.baseDir,
       humanReadable = opts.humanReadable,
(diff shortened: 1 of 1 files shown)


jantimon:
Unfortunately the build https://travis-ci.org/jfhbrook/node-ecstatic/jobs/52439466 fails because of Error: listen EADDRINUSE.

 Could you please run the tests once again?

jfhbrook:
This looks reasonable. I'll try getting to this over the weekend.

I run the tests locally anyway before merging, so you don't have to worry.

jantimon:
Okay cool thanks :+1:

PR #134, opened 2015-03-02
Fix autoIndex path (index.html) for Windows

jinjor:
I use `autoIndex` option and got an error.
```
Error: Invalid forward slash character
```
In Windows, `path.join` uses separator `\`. So I replace it with `/`.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -117,7 +117,7 @@ var ecstatic = module.exports = function (dir, options) {
 
           if (autoIndex) {
             return middleware({
-              url: path.join(encodeURIComponent(pathname), '/index.' + defaultExt)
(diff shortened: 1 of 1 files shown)


jfhbrook:
This change would break every other operating system.

jfhbrook:
Oh wait, you know what? That's a uri so this fix would be fine. But, I think we can do better and just, like, not use path.join here.

jfhbrook:
Note to self: use https://www.npmjs.com/package/url-join as appropriate.

jfhbrook:
Fixed here https://github.com/jfhbrook/node-ecstatic/commit/873aabdbf357610dea6092dc9e40cd9a693330e4

Thanks!

jfhbrook:
(Available in ecstatic@0.7.0)

PR #135, opened 2015-03-17
Add option to provide default content-type option

SirAnthony:
This patch add option to provide default content type to serve with. There is no point for text-file server to return octet-stream content, let it define its default type.


jfhbrook:
You wanna try throwing down some tests for this? Should be straightforward---the other tests use the default, so just write one that spins up a server with text/plain set as the default and assert you get what's expected.

SirAnthony:
Yes, I'll add tests & and docs in this patch a little later.

SirAnthony:
Ok, I spent 4 hours for thing I'll not use (I wrote more hardcore and simple patch for my purpose), found bug in code (mime always return something default, so charset always will be set and I'm not sure it is expected behaviour), lot of duplicated code, similar differently-written code.
I currently have no idea why this crutched test framework, who redirects standard output and throws it away leaving no opportunity to quick debugging except throwing, fails on test 3 with timeout only on node 0.10. I also have no idea what is test 3, I defined only 2 of them.
It is up to you what to do with this pull request, I can add more fixes if you suggest the way to do it, but I want to be user, not developer of this thing, and have no intention of knowing how it works.

jfhbrook:
I'm not really in a position to console you over your frustration with node-tap, but writing the test myself was easy enough.

Thanks for your contribution, it's available in ecstatic@0.7.0.

PR #138, opened 2015-03-30
Fixed double encoding in directory listings when directory name contains spaces

VilleSalonen:
If a directory listing was requested for a directory with spaces in its name, already escaped url was combined with the file names.

    Expected: subdir_with%20spaces\file_with%20space.html
    Got:      subdir_with%2520spaces\file_with%20space.html


jfhbrook:
I'm concerned about the replace being missing. Otherwise, I'll try to give this some TLC later in the week.

PR #139, opened 2015-03-31
Add ability to use multiple instances of ecstatic

### LINKED ISSUE #136: 404 on existing files in multiple instances of ecstatic
I have a setup where i feed `union` with two instances of ecstatic. I want to be able to serve two different directories as one. 

Like so: 
```js

paths = [app, tmp];

paths.forEach(function (path) {
    before.push(ecstatic({
      root: path,
      cache: cache,
      showDir: showDir,
      autoIndex: autoIndex,
      defaultExt: ext,
      handleError: false
    }));
  })
```

The problem is that the first search for a file might give an `ENOENT` and the `res.statusCode` is set to a 404. The second search gives a positive result and the the response is created for serving the file. Sadly the `statusCode` is still the 404 because it's not updated. And so the `union` server response is a 404 with the correct file. 

A quick fix would be to actively set a 200 statusCode if a file is found. This could be here : https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L142 or here https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L162

What do you think?

I could prepare a PR (with a little help on testing).

yfr:
And fixes weird issue with 404 status code.

The `res.statusCode` should always be check and adjusted in the `serve` method. Just set the `req.statusCode` in `statFile` and use it in `serve`.

Hope this makes sense. I added tests for multiple folder to serve. All test seem to work correctly.

Closes #136

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -76,26 +76,19 @@ var ecstatic = module.exports = function (dir, options) {
       fs.stat(file, function (err, stat) {
         if (err && (err.code === 'ENOENT' || err.code === 'ENOTDIR')) {
           if (req.statusCode == 404) {
-            console.log('allready 404 trying: ', req.url, req.statusCode, res.statusCode);
-            // This means we're already trying ./404.html and can not find it. So send plain text response with 404 status code
+            // This means we're already trying ./404.html and can not find it.
+            // So send plain text response with 404 status code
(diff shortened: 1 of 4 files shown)


fsmanuel:
:+1:

PR #140, opened 2015-04-14
destroy the file stream if the response closes

mafintosh:
`src.pipe(dst)` doesn't close the source stream if the destination closes. this means that ecstatic will leaks file descriptors if the response closes before the file stream ends


jfhbrook:
Oh word.

Is there any chance that destroy can throw an error or otherwise cause barf? Do we need to protect against that?

jfhbrook:
Published @0.7.3

Cheers!

PR #141, opened 2015-04-17
correct sort ordering and also applied to dirs and lolwuts

ThaNarie:
- b[0]-a[0] returns NaN, so this is never valid in a sorting function
- reverted sorting order a-z (was z-a)


jfhbrook:
Released in 0.7.4. Thanks!

PR #142, opened 2015-04-30
Prevent HTML reflection vulnerability in 500 error status handler

jre-g:
Certain IO errors containing the requested path can be echoed directly back to the user without a Content-Type header. In some cases, browsers' sniffing can result in the document being treated as HTML, allowing XSS. This is resolved by sending a Content-Type header of text/plain.

Includes a test case that can reproduce this behaviour, at least on some Linux systems. (The test should simply pass on other systems.) You can observe this test failing as expected after an intentional regression in [this Travis build](https://travis-ci.org/jre-g/node-ecstatic/jobs/60772623).


jfhbrook:
Just because I didn't merge this, doesn't mean I'm disinterested. I just hadn't gotten around to it.

jre-g:
Oops! I'm very sorry. I didn't intend to close this, or realize that I had. I was trying out a new Git client, and must have done something silly.

jfhbrook:
All good! I'll try to merge this tonight.

jfhbrook:
So I think this is a sane PR and I want to merge it, but I'm really trying to get my browser to throw up an alert and it's just not havin' it. Any hints? I'd love to be able to before/after this before clicking the magic green button.

jre-g:
I've just emailed you an example that should work. I didn't include one here because, though it probably won't make much of a difference, I didn't want to make things easier for anybody who was trying to exploit this.

jfhbrook:
Beautiful. <3

Issue #66, opened 2013-04-14
Add API entry points for custom and async mime lookup


mk-pmb:
Hi, I'm trying to serve text files with unusual filename extensions, which wind up as octet-stream due to ecstatic's reliance on the "mime" module:

```
sven:/gr2$ nodejs -e 'console.log(require("mime").lookup("common.flp"));'
application/octet-stream
```

I'd prefer them to be resolved according to my local mime types DB:

```
sven:/gr2$ file --mime common.flp
common.flp: text/plain; charset=us-ascii
```

This could be achieved by using the "mime-magic" module:

```
sven:/gr2$ nodejs -e 'var mime = require("mime-magic"); mime("common.flp", console.log);'
null 'text/plain'
```

It seems that in the "mime" module, there's currently no way to distinguish known octet-streams from fallback octet-streams ([mime bug #55](https://github.com/broofa/node-mime/issues/55)).

<del>I'd be glad if you could implement the mime-magic lookup into mainstream ecstatic.</del>

(Edit 2015-12-08: The filename-based lookup is fine for default. If mime-magic lookup ever gets into mainstream ecstatic, only use it as a fallback for when filename-based mime has no idea.)

dotnetCarpenter:
I would really like this in [http-server](https://github.com/indexzero/http-server), so guess this is the *upstream* repo to follow. It's a (**REALLY**) nice-to-have not a need-to-have, yet.

mk-pmb:
FYI, I had tried tackling this again a few days ago, but couldn't get the tests to pass for current master (3.0.0), even with env proxy config adjusted to no proxy. Maybe the non-proxy parts of #180 can help fix them so anyone interested in this, consider fixing that one, too.

mk-pmb:
I'd like to change our dependency on the old `mime@^1` to the `mundane-mime-v1` proxy package I just published, so we can still use v1 for people who like to `.load()` Apache style MIME type lists, while at the same time allowing partial migration to the latest `mime`. Yes/no?

jfhbrook:
Oh, to better enable replacing the dependency in a scenario where npm dedupes?

mk-pmb:
Actually it was just the easiest way to make both versions live side by side on my machine without npm going on a rampage, enforcing default whitespace style in hundreds of `package.json`s even when no values were changed. However, now considering your question, it might indeed have been a side effect of a dedupe attempt.

mk-pmb:
[...]

PR #143, opened 2015-05-11
Mimetype

dotnetCarpenter:
Look story short: I really like [The Mother of All Demos](http://en.wikipedia.org/wiki/The_Mother_of_All_Demos) and looked into the modern version of [HyperScope](http://hyperscope.org/). Sadly, it is already based on ancient tech, so wanted to bring it up to date before it is too difficult. First choice was to use [http-server](https://github.com/indexzero/http-server). Unfortunately the .opml files HyperScope uses didn't load because of the mime-type/content-type and http-server has no way of defining custom types.

So I gave it try, fixing #66 and everything looks good but a few things, that I hope you guys can help with.
1. Firstly I'm on windows right now, so get a lot of line endings equality errors (`\r\n` instead of `\n`) when I run the test suite. I think everything passes but it is difficult to see.
2. I'm not sure how to write out warnings. In this case, when a path to a `.types` file is wrong (404). I would like to inform the user, not crash `estatic` or ignore the error.
3. How do I go about testing number 2?

Anyway, I hope you look positive at this feature and perhaps I can reach my end goal :+1:


jfhbrook:
ecstatic doesn't have a way of emitting warnings or other log events, and I think that's out of the scope of this PR--though perhaps something I should consider! Given the current state and where you're doing the check, I would throw.

dotnetCarpenter:
All right, I'll throw. Should I squash all the commits into 1? Or anything else?

jfhbrook:
Ehh, I can manage squashing if I think it's necessary. In general I don't get overly finnicky about commit history of outside contributions because it would take way too much time.

dotnetCarpenter:
Cool. So I'm almost there. Got an issue with the content-type.js test case. My `teardown` function doesn't properly tear down ecstatic. If you comment out line `63: 'mime-types': 'custom_mime_type.types'` the test still passes because the mime-type for .opml is defined in the previous test. Any quick ideas to what would kill the ecstatic instance?

jfhbrook:
You'll see that in a bunch of the tests I have timeout code that kills ecstatic if all the requests are done but a mystery open handle is still kicking around. I don't remember why I thought this was okay, but if it's good enough for the other tests...

dotnetCarpenter:
I've asked for a reset implementation in `mime` of mime-types. https://github.com/broofa/node-mime/issues/124

I'll try to spawn ecstatic in child process and then kill that between tests.

dotnetCarpenter:
Hmm.. since I can't setup forks without support in ecstatic.js, which I rather not do just for testing and that the code we are testing passes each individual  test in content-type.js, I'm inclined to let it go for now.
Interestingly, it seem that node [cluster](https://nodejs.org/api/cluster.html) could be used. But t's out of scope for this PR.

dotnetCarpenter:
@jfhbrook if I use setTimeout process.exit(0), then I would have to split `content-type.js` into 3 different files, right?

jfhbrook:
I don't think so? You should be able to handle the timeout at the same place as you call `server.close`.

As an aside though, I don't think we have to test the mime package's behavior. We should assume it behaves as-advertised.

jfhbrook:
I appreciate all the work you're putting into this by the way. Looks like good work!

dotnetCarpenter:
I think it is good practice to test the parts of a third-party package that you use in your project. This automates detection of changes made in that library. It should make it very comforting to upgrade the package. If anything changes, we will know before pushing a new version with updated dependencies. I understand your stand point though. But I needed to learn `mime` anyway, so might as well do that in a unit test.

I'll give it a shot with timeout....

dotnetCarpenter:
Nothing can happen after `process.exit(0)`. So have to split up the test into 3 different files. In the long run, ecstatic should be able to proper clean up after itself. In long running processes or memory constraint devices, this could be a real issue.

dotnetCarpenter:
I can't think of anything else to do in this PR. Ready for merge?

dotnetCarpenter:
@jfhbrook wait. While implementing indexzero/http-server#35 I realized that the path to the .types file should be independent of the root path. Say I want to serve files from one directory but want to load mime-types from another directory.

dotnetCarpenter:
The README.md says that default for root is `root          : __dirname + '/public',` so I made the following condition in ecstatic.js:
```js
var root = dir && path.join(path.resolve(dir), '/') || path.resolve(__dirname + 'public'),
```

Is that fine?

dotnetCarpenter:
After working a bit with the default path to public, I've changed my mind. I find it confusing. Much better to have the user define a root.
I'm trying to make the test, platform agnostic so I can actually run all of them on windows. And tripped over [line 23 in common-cases](https://github.com/jfhbrook/node-ecstatic/blob/master/test/common-cases.js#L23). Surely it should be **c.js** as it is the content of `d.js`. it's a bit worrying that the test has passed so far. But my current experience is that `ecstatic` isn't that easy to test...

dotnetCarpenter:
So I went through all of the tests and only one is failing. It turns out that many of the tests didn't run - at least I don't think so.
[Line 100 in range.js](https://github.com/jfhbrook/node-ecstatic/blob/master/test/range.js#L100) expects **3-11/12** but gets **3-10/11**. What is the correct value?

dotnetCarpenter:
@jfhbrook I've update tap to version 1.0 and with that I had to move the scripts in the test folder that is not tap tests into a seperate folder. I already had a *secret* folder to test for .types file being in another directory that the root (it had to be outside public), so I used that.

dotnetCarpenter:
@jfhbrook do you know what [malformed-dir.js](https://github.com/dotnetCarpenter/node-ecstatic/blob/mimetype/test/malformed-dir.js) and [malformed.js](https://github.com/dotnetCarpenter/node-ecstatic/blob/mimetype/test/malformed.js) is suppose to test? I'm having trouble getting them to pass and don't know what they are suppose to do.

dotnetCarpenter:
@jfhbrook ok, I get that they test what happens with a malformed path. I couldn't, for the life of me, get [http.get](https://nodejs.org/api/http.html#http_http_get_options_callback) to work. So changed it to use request as in many of the other tests.

We're still some tests short. I could really need a little feedback....

dotnetCarpenter:
right, so before we had 267 tests and now we have 255. It turns out that the old version of `tap` counted none test cases as well, i.e. `common-cases.js` counted 2 tests but has none and `304.js` counted 3 but only has 2.
So if travis agrees - I think I'm finally done with this PR (by far the longest I've ever made)

jfhbrook:
Yeah, okay. I'll try to review some time this week.

dotnetCarpenter:
@jfhbrook thanks. I know it's mouthful - it was for me - and will do what I can to help ease the merging of this. In the mean time, I'll continue work on indexzero/http-server#35, based off this PR.

jfhbrook:
You know ecstatic can be ran as a bin, right? `npm install ecstatic -g && ecstatic`

dotnetCarpenter:
@jfhbrook eh, yes. What do you mean?

jfhbrook:
I guess what I'm saying is, I don't understand what value http-server adds.

jfhbrook:
ehh nvm, it looks like http-server ships with cors support and a few other things.

dotnetCarpenter:
ah. I like http-server it super easy to use and I only knew about http-server, not ecstatic. I came to ecstatic because I wanted custom mimetypes. And since http-server depend on ecstatic... well, the rest [is history](#issue-75282835). - yep it started here :)

README.md
```diff
@@ -1,4 +1,4 @@
-# Ecstatic [![build status](https://secure.travis-ci.org/jfhbrook/node-ecstatic.png)](http://travis-ci.org/jfhbrook/node-ecstatic)
+# Ecstatic [![build status](https://secure.travis-ci.org/jfhbrook/node-ecstatic.png)](http://travis-ci.org/jfhbrook/node-ecstatic) [![dependencies status](https://david-dm.org/jfhbrook/node-ecstatic.svg)](https://david-dm.org/jfhbrook/node-ecstatic)
```
jfhbrook: What is this?

ChangeLog.md
```diff
@@ -1,3 +1,12 @@
+2015/05/13 Version 0.8.2
+- Updated dev depedencies and fixed all failing tests
+
+2015/05/12 Version 0.8.1
+- Fixed an issue where .types file is in another directory that the root directory
+
+2015/05/11 Version 0.8.0
```
jfhbrook: I should probably add something about the changelog in the contributing guidelines---I like to manage this myself while bumping and publishing releases. But the notes here are really helpful! I'll just end up munging this a bit.

lib/ecstatic/opts.js
```diff
@@ -10,15 +10,20 @@ module.exports = function (opts) {
       gzip = false,
       defaultExt = '.html',
       handleError = true,
-      serverHeader = true;
-      contentType = 'application/octet-stream';
+      serverHeader = true,
+      contentType = 'application/octet-stream',
+      mimeTypes;
+
+  function isDeclared(k) {
```
jfhbrook: Pulling this out is probably a good idea. I approve.

package.json
```diff
@@ -2,7 +2,7 @@
   "author": "Joshua Holbrook <josh@nodejitsu.com> (http://jesusabdullah.net)",
```
jfhbrook: Note to self: change email address

test/escaping.js
```diff
@@ -1,30 +1,19 @@
...
  
   server.listen(0, function () {
     var port = server.address().port;
     request.get('http://localhost:' + port + "/curimit%40gmail.com%20(40%25)", function (err, res, body) {
       t.ifError(err);
       t.equal(res.statusCode, 200);
-      t.equal(body, 'index!!!\n');
-      t.end();
+      t.equal(eol.lf(body), 'index!!!\n');
+      server.close(function() { t.end(); });
```
jfhbrook: Does this fix the open handles issue?

test/express-error.js
```diff
@@ -12,7 +12,7 @@ var root = __dirname + '/public',
 
 mkdirp.sync(root + '/emptyDir');
 
-var cases = require('./common-cases-error');
+var cases = require('./secret/common-cases-error');
```
jfhbrook: Is the name of this folder something dictated by tap@1.x or does it just ignore any subfolders?

test/range.js
```diff
@@ -95,9 +97,14 @@ test('partial range', function (t) {
     request.get(opts, function (err, res, body) {
       t.ifError(err);
       t.equal(res.statusCode, 206, 'partial content status code');
-      t.equal(body, 'e!!</b>\n');
+      t.equal(eol.lf(body), 'e!!</b>\n');
```
jfhbrook: eol fixes windows problems?

test/range.js
```diff
@@ -95,9 +97,14 @@ test('partial range', function (t) {
...
       t.equal(parseInt(res.headers['content-length']), body.length);
-      t.equal(res.headers['content-range'], 'bytes 3-10/11');
+      
+      if (process.platform === 'win32') {
+        t.equal(res.headers['content-range'], 'bytes 3-11/12');
```
jfhbrook: This mismatch is due to `\r`'s?

README.md
dotnetCarpenter: What *huge* issue I had, was updating from tap 0.4 to 1.0. Since I spend a few hours doing that. I though a badge was in order :). It a free service that tells us about the health of our dependencies. See here: https://david-dm.org/jfhbrook/node-ecstatic#info=devDependencies&view=table (and don't worry - it will look better when you merge this PR)

package.json
dotnetCarpenter: yeah, it made me think that you guys sat together and made http-server and ecstatic. I even went to freenode.org#nodejitsu and asked around for you a few times.

test/escaping.js
dotnetCarpenter: yes, as far as I can tell. I just switch the logic around. Instead of closing the server on t.end, I call t.end() when the server is closed.

test/express-error.js
dotnetCarpenter: [apparently we were using it wrong](https://github.com/isaacs/node-tap/issues/128#issuecomment-101453724). tap does not expect any files that are not tap test files. So I moved the non-test files into secret/ and use the following glob pattern: **test/*.js**

test/range.js
dotnetCarpenter: yep. On windows new line is \r\n or CRLF. On *nix it is \n or LF. I've also heard that ancient OS's uses \r OR CR. But I've never seen it IRL.

test/range.js
dotnetCarpenter: yep, that windows.

ChangeLog.md
dotnetCarpenter: Do as you like. I put them there for you

dotnetCarpenter:
@jfhbrook should mention that union-multiple-folders.js creates a [object global] folder in the root. I haven't checked why but pretty sure that that didn't happen before tap@1.0

dotnetCarpenter:
hmm trying to figure out what @yfr wanted to do in union-multiple-folders.js. I'm not sure that test ever made sense. Did I delete something?

README.md
```diff
@@ -139,6 +141,13 @@ on all responses served by ecstatic.
 Set `opts.contentType` in order to change default Content-Type header value.
 Defaults to **application/octet-stream**.
 
+### `opts.mimeTypes`
+
+Add new or override one or more mime-types. This affects the HTTP Content-Type header.
+Can either be a path to a [`.types`](http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types) file or an object hash of type(s).
+    
+    ecstatic({ mimeType: { 'mime-type': ['file_extension', 'file_extension'] } })
```
dotnetCarpenter: Does this kins of doc make sense? An example would be `ecstatic({ mimeType: { 'node/ecstatic+fun': ['ecs', 'fun'] } })` to get `content-type: node/ecstatic+fun; charset=UTF-8` with `.esc` and `.fun` files.

lib/ecstatic.js
```diff
@@ -31,6 +31,17 @@ var ecstatic = module.exports = function (dir, options) {
   opts.root = dir;
   if (defaultExt && /^\./.test(defaultExt)) defaultExt = defaultExt.replace(/^\./, '');
 
+  // Support hashes and .types files in mimeTypes @since 0.8
+  if (opts.mimeTypes) {
+    if (typeof opts.mimeTypes === 'string') {
+      //TODO: should handleError have any effect here?
```
dotnetCarpenter: @jfhbrook what do you think about the handleError option? Should we not throw if it is true?

README.md
jfhbrook: yeah, I think that makes sense.

lib/ecstatic.js
jfhbrook: Yes, I think throwing is still appropriate.

dotnetCarpenter:
@jfhbrook thanks for your time. Is there anything I can do to help this PR getting merged? I'm off-line this weekend but will be back mid next week.

test/express-error.js
dotnetCarpenter: Or to actually answer your question: Yes, tap ignores sub folders, if your glob pattern does.

jfhbrook:
Released in https://github.com/jfhbrook/node-ecstatic/releases/tag/0.8.0

Thanks!

dotnetCarpenter:
@jfhbrook that's awesome  - thanks!

_minor detail:_ though `public` is a _fixture_ folder 8b27bba

PR #145, opened 2015-07-09
Weak etags

mofojed:
Add support to generate weak etags instead of strong etags.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -218,12 +218,7 @@ var ecstatic = module.exports = function (dir, options) {
       res.setHeader('cache-control', cache);
 
       // Return a 304 if necessary
-      if ( req.headers
-        && (
-          (req.headers['if-none-match'] === etag(stat, weakEtags))
-          || (new Date(Date.parse(req.headers['if-modified-since'])) >= stat.mtime)
-        )
-      ) {
+      if (shouldReturn304(req, stat)) {
         return status[304](res, next);
       }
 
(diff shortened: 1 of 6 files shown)


dotnetCarpenter:
@mofojed I just looked up weak etags (I had never heard about them before) on wikipedia and I still don't get it. Why are they needed?

https://en.wikipedia.org/wiki/HTTP_ETag#Strong_and_weak_validation

jfhbrook:
My tags are STRONG like BEAR

jfhbrook:
Seriously though: Can you explain the use case for this? I'm not actually an expert in client-side caching so I'll need a little help with this one.

mofojed:
Weak etags are used when the response is not necessarily byte-for-byte equivalent, but can replace the existing contents.

This comes in handy when gzipping content, as the gzipped content is not necessarily byte-for-byte. When behind a proxy like nginx, it automatically converts strong etags to weak etags when gzipping content, which is correct. However, then node-ecstatic can't match the etag, since it's stored the strong etag.
By using weak etags instead of strong etags node-ecstatic can properly serve up 304's for gzipped content when behind an nginx proxy :)

killswitch:
> My tags are STRONG like BEAR
:+1:

dotnetCarpenter:
Sounds good to me and you made test. But isn't it a tad _weak_ that we only add a `W/` to the header?  I can't see where you change the actual etag value. But perhaps I'm missing something?

mofojed:
Prefixing the tag with `W/` is all that's needed to convert a strong etag to a weak etag: https://en.wikipedia.org/wiki/HTTP_ETag#Strong_and_weak_validation

This is the same conversion that nginx does implicitly when set up as a proxy:
http://mailman.nginx.org/pipermail/nginx-announce/2014/000141.html:
```    *) Feature: weak entity tags are now preserved on response
       modifications, and strong ones are changed to weak.```

**Comment by mofojed** [comment]
Actually - upon re-reading RFC2616, section 13.3.3 on strong/weak validators (here: https://www.ietf.org/rfc/rfc2616.txt) I think I'm going about this incorrectly. I think it's just weak comparison that needs to be implemented, eg:
```
      - The weak comparison function: in order to be considered equal,
        both validators MUST be identical in every way, but either or
        both of them MAY be tagged as "weak" without affecting the
        result.
```

Let me revise the change.

**Comment by mofojed** [comment]
Adding a change that correctly performs strong vs. weak comparison for etag validators. Please take a look, and check out the spec I've referenced.

Any questions welcome :)

**Comment by dotnetCarpenter** [comment]
Code looks good but I need to read the specs... It's on my todo

**Comment by dotnetCarpenter** [comment]
*update* reading the specs...

File: lib/ecstatic.js [hunk 1]
```diff
@@ -245,6 +241,39 @@ var ecstatic = module.exports = function (dir, options) {
...
+        // Client did not provide any conditional caching headers
+        return false;
+      }
+
+      // If any of the headers provided don't match, then don't return 304
+      if (modifiedSince && (new Date(Date.parse(modifiedSince))) < stat.mtime) {
+        return false;
+      }
+
+      if (clientEtag) {
```
jfhbrook: I think this check is extraneous because you already checked that there's *a* header and it's not modifiedSince

jfhbrook:
Based on my wikipedia-grade understanding of weak etags and having read through this, beyond my one comment I think this looks good. I'm gonna give @dotnetCarpenter a chance to read the spec though.

lib/ecstatic.js
mofojed: I believe this check is still required. At this point, there's either no modifiedSince header OR a modifiedSince header that falls within the proper date range. We don't know that the clientEtag header exists for sure.

jfhbrook: Okay yeah, I follow.

jfhbrook:
What's the worst that can happen, right? :D

Thanks.

PR #147, opened 2015-08-15
Icons and CSS

fenwick67:
Added icons and CSS to the directory listing, with some corresponding tests.  This is what it looks like:
image

If the option `icons` is set to `true`, it will add some `<i>` elements with a class attribute based on the file extension.  The icons are set with CSS based on this class.  `icons` is `false` by default.

This involved adding a directory to `/lib` called `_ecstatic-assets` and serving files out of it if `showDir` is true.  Note this folder and its assets aren't shown in the directory listing.  Also this is not indexed off of `baseDir`, it is always off of the root (but I can change this).

I also removed the inline styles on the table elements on the directory listing, opting for CSS instead.  This also opens the lib up for more styling, if somebody wanted to do that.  You could also theoretically put some js or other assets into `_ecstatic-assets`.

One outstanding issue, it returns a 500 error if you visit `/_ecstatic-assets` or the `/_ecstatic-assets/icons` subdirectory (the folders themselves).  I figured it makes sense not to serve the directory listing in these cases, but I'm not sure how to get it to show a 404 instead of 500.


fenwick67:
@jfhbrook any thoughts?

jfhbrook:
oh whaaat I did NOT know this was here. I haven't done a review or anything but on the face of it this sounds dope.

jfhbrook:
uhhh what's the licensing on these icons?

lib/ecstatic.js
```diff
@@ -57,27 +57,39 @@ var ecstatic = module.exports = function (dir, options) {
...
     }
-
-    var file = path.normalize(
-          path.join(root,
-            path.relative(
-              path.join('/', baseDir),
-              pathname
+    
+    // try and serve assets
+    var assetsDirectory = '_ecstatic-assets'
```
jfhbrook: I don't like the shadowing of the directory structure that this does, so it at least has to be possible to turn off. I think the option that turns on icons should also turns on the css---maybe `pretty` is a better name?

lib/ecstatic/showdir.js
```diff
@@ -108,6 +109,8 @@ module.exports = function (opts, stat) {
             '  <head>',
             '    <meta charset="utf-8">',
             '    <meta name="viewport" content="width=device-width">',
+            '    <link href="/_ecstatic-assets/ecstatic.css" rel="stylesheet" type="text/css" />',
```
jfhbrook: An alternative to feature-flagging all CSS: embed it directly. That could be cool.

lib/ecstatic/showdir.js
```diff
@@ -147,6 +151,8 @@ module.exports = function (opts, stat) {
             '/ <a href="https://github.com/jesusabdullah/node-ecstatic">ecstatic</a> ' +
             'server running @ ' +
             he.encode(req.headers.host || '') + '</address>\n' +
+            (icons ? '<a href="http://www.famfamfam.com/lab/icons/silk/">Silk Icons</a>':'')
```
jfhbrook: this should only show up if icons are on

jfhbrook: Oh nevermind, this is ternary---should be fine

lib/ecstatic.js
jfhbrook: If it only serves up icons and you can opt out (or in), we could just as soon put this in `/icons`

jfhbrook:
So the 500 occurs when it tries to do a createReadStream on a folder?

lib/ecstatic/showdir.js
fenwick67: You mean put it in a <`style>` tag?

lib/ecstatic.js
fenwick67: I will change this directory to something like `/_icons`, and make it so that it won't get served up unless the option is on.  I'll embed the non-icon page styles directly in a `<style>` tag since there are only 2 rules.

lib/ecstatic/showdir.js
jfhbrook: Yeah, just <style> tag all the css like a boss

fenwick67:
Licensing for the icons is CC BY 2.5, I need to add a link to the license somewhere... not 100% sure where to put that (readme.md ?).  Will investigate what to do about the 500 too.

lib/ecstatic.js
jfhbrook: Do you think those lines would look good along with the http style status pages in https://github.com/jfhbrook/node-ecstatic/pull/151/files ? I might try to copypasta this a little.

jfhbrook:
Closing in favor of https://github.com/jfhbrook/node-ecstatic/pull/188 . Your name will still be in the contributors file, and hey, feedback appreciated!

Issue #119, opened 2014-08-29
fall back to PORT env var


bmeck:
if no explicit option is set for the port to listen on but PORT has a sane value ideally it would be listened to

jfhbrook:
We have this now.

PR #150, opened 2015-08-25
added fallback to process.env.PORT

jekrb:
As mentioned in #119, this allows the server to listen to the port given in process.env.PORT if the number is of sane value.
If no value is given in process.env.PORT or the port number is _not_ sane then the server listens on
``` opts.port || opts.p || 8000``` as usual.

*I chose sanity to be ports >1024 and =< 65536 based off [this list of well-know ports](http://www.webopedia.com/quick_ref/portnumbers.asp)

I wasn't really sure how to write test for this.  It seemed like an easy change but if test are needed I could see what I can do.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -282,7 +282,7 @@ function decodePathname(pathname) {
 if (!module.parent) {
   var http = require('http'),
       opts = require('minimist')(process.argv.slice(2)),
-      envPORT = process.env.PORT,
+      envPORT = parseInt(process.env.PORT, 10),
       port = envPORT > 1024 && envPORT <= 65536 ? envPORT : opts.port || opts.p || 8000,
       dir = opts.root || opts._[0] || process.cwd();
 
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -282,7 +282,8 @@ function decodePathname(pathname) {
 if (!module.parent) {
   var http = require('http'),
       opts = require('minimist')(process.argv.slice(2)),
-      port = opts.port || opts.p || 8000,
+      envPORT = process.env.PORT,
+      port = envPORT > 1024 && envPORT <= 65536 ? envPORT : opts.port || opts.p || 8000,
       dir = opts.root || opts._[0] || process.cwd();
 
   if (opts.help || opts.h) {
--- a/test/process-env-port.js
+++ b/test/process-env-port.js
@@ -1,0 +1,48 @@
+var test = require('tap').test,
+    ecstatic = require('../lib/ecstatic'),
+    spawn = require('child_process').spawn
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by jfhbrook** [comment]
Testing isn't bad I think, you just spin up ecstatic as a child process and curl the endpoint :D

**Comment by jekrb** [comment]
Added test.  To be honest, I'm new to using child_process and tap.  I'm open to criticism if this looks like the wrong approach. :)

**Comment by jfhbrook** [comment]
Right idea on the tests, though you could use the request module instead of spawn('curl' ---I *believe* it's already in the devDeps

(easy to fix though)

**Comment by jekrb** [comment]
Good call, I've committed the refactored test

File: test/process-env-port.js [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+  ecstatic.stdout.on('data', function (data) {
+    t.pass('ecstatic should be started')
+    checkServerIsRunning('http://0.0.0.0:' + process.env.PORT, ecstatic, t)
+  })
+})
+
+insanePorts.forEach(function (port) {
+  test('insane port: ' + port, function (t) {
+    t.plan(2)
+    process.env.PORT = port
```
jfhbrook: You can pass this into spawn as an argument---something like `{ env: { PORT: port }}` (we'd have to hit up the docs to be sure on that)

lib/ecstatic.js
```diff
@@ -282,7 +282,8 @@ function decodePathname(pathname) {
 if (!module.parent) {
   var http = require('http'),
       opts = require('minimist')(process.argv.slice(2)),
-      port = opts.port || opts.p || 8000,
+      envPORT = process.env.PORT,
```
jfhbrook: I think this will break without a parseInt(process.env.PORT, 10)

jekrb:
> You can pass this into spawn as an argument---something like { env: { PORT: port }} (we'd have to hit up the docs to be sure on that)

Cool, I'll make that change.

> I think this will break without a parseInt(process.env.PORT, 10)

Since ports can only be 16-bit integers, it seems like ecstatic, (or node, or unix) handles this already.  Giving process.env.PORT something like 9090.9 will start the server on 9090.  I'll definitely add it to be sure, this was just something I observed.

jekrb:
Thanks for the recommend changes.
Also, I added an extra test to be sure floating point decimals are handled.

jekrb:
There's a pretty reusable pattern in the test.  I'm going to clean that up real quick.

jekrb:
Also, thank you for encouraging me to write the test.  I feel it was a healthy learning process. Regardless if any of this pr actually adds value to your project, the iterations were valuable to me.

jfhbrook:
Thanks!

PR #151, opened 2015-08-27
Potential XSS mitigations and encoding corrections

jre-g:
The change I submitted in #142 was adequate to prevent XSS in Firefox, but potentially not in all browsers. Previous versions of Internet Explorer have aggressively sniffed `text/plain` content as `text/html` despite the header, and it may be possible to provoke this behaviour in current versions if compatibility modes are enabled. Bugs have also resulted in this behaviour in other browsers in some contexts, as recently as iOS 6 Safari. I'm not certain whether it's possible to trigger this for these specific error messages (I don't have a proof of concept), but in the interest of taking a conservative approach to security, I suggest replacing the `text/plain` responses with unencoded bodies with `text/html` responses with encoded bodies.

This PR also corrects two other locations where data was being outputted without being encoded. There are *probably* not dangerous, but it would be more correct to encode them and avoid creating non-validating documents.

Please see [here](https://travis-ci.org/jre-g/node-ecstatic/jobs/77593226) for a Travis build with the test changes but without the other changes, demonstrating that regressions cause the tests to fail.

---

Encoding current directory name in directory listings. HTML in directory names (unlikely but possible) could previously have been rendered in the page.

Encoding query strings in subdirectory links in directory listings. It is unclear if the query string values could actually have been rendered in any cases, but the unencoded values could produce inconsequentially incorrect links (e.g. `/?&gt;` would produce subdirectory links like `/dir/>`) and cause HTML validation to fail.

Replacing text/plain responses messages for 400 and 500 errors with text/html response pages. This should not be neccessary in theory, but older versions of some browsers including Internet Explorer (and possibly newer versions in certain contexts) sometimes render text/plain content as text/html for backwards compatibility with poorly-configured servers, even if `X-Content-Type-Options: nosniff` is specified. Since the errors messages in these handlers could potentially include content reflected from the requests, the most conservative and safe approach is to use an HTML response with the error messages encoded.

Also updates LICENSE.txt to include the current year and some additional contributors.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
-Copyright (c) 2013 Joshua Holbrook
+Copyright (c) 2013-2015 Joshua Holbrook, Jon Ege Ronnenberg, James Halliday,
+and Google Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/lib/ecstatic/showdir.js
+++ b/lib/ecstatic/showdir.js
@@ -108,10 +108,10 @@ module.exports = function (opts, stat) {
             '  <head>',
             '    <meta charset="utf-8">',
             '    <meta name="viewport" content="width=device-width">',
-            '    <title>Index of ' + pathname +'</title>',
+            '    <title>Index of ' + he.encode(pathname) +'</title>',
             '  </head>',
             '  <body>',
-            '<h1>Index of ' + pathname + '</h1>'
+            '<h1>Index of ' + he.encode(pathname) + '</h1>'
           ].join('\n') + '\n';
 
           html += '<table>';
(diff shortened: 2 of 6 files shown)


LICENSE.txt
```diff
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
-Copyright (c) 2013 Joshua Holbrook
+Copyright (c) 2013-2015 Joshua Holbrook, Jon Ege Ronnenberg, James Halliday,
+and Google Inc.
```
jfhbrook: google?

jre-g: I'm making this PR for my work as a Google employee. It was requested that I add this to the copyright notice. As I personally understand things (I am not a lawyer), Google owns the copyright to the few lines being added. I added the largest contributors listed on GitHub to reflect their ownership as well.

jfhbrook:
Thanks!

PR #153, opened 2015-09-17
-H/--header like curl

substack:
This patch adds `-H` to the `ecstatic` command, so if you want to send a CORS header on every request, you could for example do:

```
ecstatic -p 8001 . -H 'Access-Control-Allow-Origin: *'
```

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -310,7 +310,7 @@ function decodePathname(pathname) {
 
 if (!module.parent) {
   var http = require('http'),
-      opts = require('minimist')(process.argv.slice(2)),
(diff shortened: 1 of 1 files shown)


jfhbrook:
In general, I think functionality should be in the core middleware, and CLI arguments should just proxy to it.

Regarding default headers: I have a little resistance to this only because ecstatic already has 30k features, but I think being able to do this is useful and I don't see any reasonable alternative outside the module for being able to do so.

jfhbrook:
As far as going with args conventions as in curl, that sounds sane enough.

substack:
closed in favor of #158.

PR #155, opened 2015-09-22
Fix links for filenames with special characters

clee:
I happened to have a file with a '#' in the filename, and when using @indexzero's http-server (which uses ecstatic for autoIndex) the filename wasn't being properly encoded. This should fix that (as well as any files which use "; , / ? : @ & = + $" in the filename as well).


clee:
Oops, I meant showDir, not autoIndex. And it looks like there's a test that fails, but because it expects the old behavior (which I think is broken).

PR #127, opened 2015-02-03
encodeURI instead of encodeURIComponent

vvo:
Hrefs were malformed in the generated index.html, added a test

PR #156, opened 2015-09-22
Use encodeuricomponent

clee:
I happened to have a file with a '#' in the filename, and when using @indexzero's http-server (which uses ecstatic for autoIndex) the filename wasn't being properly encoded. This should fix that (as well as any files which use "; , / ? : @ & = + $" in the filename as well). Also updated a test that broke with the change.

### DIFF
--- a/lib/ecstatic/showdir.js
+++ b/lib/ecstatic/showdir.js
@@ -121,7 +121,7 @@ module.exports = function (opts, stat) {
           var writeRow = function (file, i) {
             // render a row given a [name, stat] tuple
             var isDir = file[1].isDirectory && file[1].isDirectory();
-            var href = parsed.pathname.replace(/\/$/, '') + '/' + encodeURI(file[0]);
+            var href = parsed.pathname.replace(/\/$/, '') + '/' + encodeURIComponent(file[0]);
 
             // append trailing slash and query for dir entry
             if (isDir) {
--- a/test/showdir-href-encoding.js
+++ b/test/showdir-href-encoding.js
@@ -25,7 +25,7 @@ test('url encoding in href', function (t) {
     request.get({
       uri: uri
     }, function(err, res, body) {
-      t.match(body, /href="\/base\/show\-dir\-href\-encoding\/aname\+aplus\.txt"/, 'We found the right href');
+      t.match(body, /href="\/base\/show\-dir\-href\-encoding\/aname%2Baplus\.txt"/, 'We found the right href');
       server.close();
       t.end();
     });


jfhbrook:
Looks legit. I gotta walk through this though cause this is ringing bells.

Going through the git blame on test/showdir-href-encoding it looks like this test was adjusted to specifically check for proper uri encoding https://github.com/jfhbrook/node-ecstatic/commit/4f637104549f713302546ce03e46737ee5660d57 but it also looks like I was at one time I was also double-encoding everything but the file https://github.com/jfhbrook/node-ecstatic/commit/6c211a3afa0edb6a745db488fcf2e35a6981bc3b

I guess what I'm getting at is: We're sure this is correct? I'd hate to break something that was fixed in a previous PR.

jfhbrook:
This is the PR in question https://github.com/jfhbrook/node-ecstatic/pull/127

jfhbrook:
I dug into this a little. I *think* your fix is correct. It *looks* like, while + doesn't *need* to be encoded, that it *can* be encoded without much harm.

Given that, we should probably make the test more robust and include spesh characters in both the folder and the file? We should also check that clicking the link works (I'm guessing you did that?)

indexzero:
@jfhbrook I agree with you. I think this is correct as well. According to the RFC mentioned in #127 (see: http://tools.ietf.org/html/rfc1738) all unsafe characters should be encoded:

>  Unsafe:
>
>   Characters can be unsafe for a number of reasons.  The space
>   character is unsafe because significant spaces may disappear and
>   insignificant spaces may be introduced when URLs are transcribed or
>   typeset or subjected to the treatment of word-processing programs.
>   The characters "<" and ">" are unsafe because they are used as the
>   delimiters around URLs in free text; the quote mark (""") is used to
>   delimit URLs in some systems.  The character "#" is unsafe and should
>   always be encoded because it is used in World Wide Web and in other
>   systems to delimit a URL from a fragment/anchor identifier that might
>   follow it.  The character "%" is unsafe because it is used for
>   encodings of other characters.  Other characters are unsafe because
>   gateways and other transport agents are known to sometimes modify
>   such characters. These characters are "{", "}", "|", "\", "^", "~",
>   "[", "]", and "`".
>
>   All unsafe characters must always be encoded within a URL. For
>   example, the character "#" must be encoded within URLs even in
>   systems that do not normally deal with fragment or anchor
>   identifiers, so that if the URL is copied into another system that
>   does use them, it will not be necessary to change the URL encoding.
> 

Oddly enough even though #127 referenced this RFC and then did not follow it correctly.

jfhbrook:
Okay, I'm with it. Let's update the test to use a folder with special characters, sanity test manually and get this out.

jfhbrook:
This'll be published as v1.0.1 as soon as I can get npm to actually let me publish

clee:
@jfhbrook Yes, I tested this change with files including both + and # symbols (as well as a handful of others, but not every single special char) and in all the cases I tried, the old version failed to serve the files properly and this version worked as I expected. :)

derhuerst:
@indexzero I think the RFC sounds a lot like the web server should *not* try to encode duplicate number signs (`#`), but the *client*.

> All unsafe characters must always be encoded within a URL. For
example, the character "#" must be encoded within URLs even in
systems that do not normally deal with fragment or anchor
identifiers, so that if the URL is copied into another system that
does use them, it will not be necessary to change the URL encoding.

Also, this would not conform with the [path definition in RFC 3986](http://tools.ietf.org/html/rfc3986#section-3.3), which states the following.

> The path is terminated by the first question mark ("?") or number sign ("#") character, or by the end of the URI.

clee:
@derhuerst I think your interpretation of the RFCs is incorrect here. If I have a file named `"Book #1.epub"`, I do not ever want a web server to attempt to serve me a file named `"Book "` with a fragment identifier of `1.epub`. I want it to serve me the file with the special characters properly escaped. And that's how I read the RFC1738 text; you MUST escape these characters, even on systems that DON'T support fragment identifiers. There's no conflict with RFC 3986; they're saying you DON'T want your path to end at the first '#' so you have to encode it to get the proper behavior.

derhuerst:
@clee 

Your point makes sense, but the RFC clearly says "terminated by the *first* question mark or number sign character". Maybe it's just impossible to serve files with number signs.

derhuerst:
Do you know how larger web servers deal with it? nginx? Apache?

clee:
If you have a file with a `#` in the filename and you enable autoindex on both Apache and nginx, it correctly encodes the `#` as `%23` just like this patch does. It is not impossible to serve files with number signs. You just have to encode the filenames with encodeURIComponent.

derhuerst:
@clee Firefox, curl and the node `url` module seem to split the URL correctly, complying with RFC 3986 (correctly IMHO). So when does the problem mentioned above even occur?

clee:
@derhuerst 

* Try `npm install http-server` and then revert this code to the previous version (not using encodeURIComponent on file[0]) in `node_modules/http-server/node_modules/ecstatic/lib/ecstatic/showdir.js`
* Run `http-server` in a directory containing multiple files with a `#` in their actual filename.
* Using Firefox or curl, attempt download the files with the `#` in their filenames.

It won't work, because without using encodeURIComponent on file[0], the URI pointing at the file will get parsed into an incomplete filename, a `#` sign, and then a nonexistent URI fragment.

Using encodeURIComponent allows the full filename to actually be encoded into the URI, which is exactly what we want.

clee:
@derhuerst For reference, the difference I see with a file named `Book #1.epub` is:

* With `encodeURI`, the link generated is `<a href="/Book%20#1.epub">`
* With `encodeURIComponent`, the link generated is `<a href="/Book%20%231.epub">`

You see why the first link is non-functional, correct? And why the second link is the one we actually want? (Aside from the fact that the second link is the way that Apache and nginx both encode the same filename in their autoindex code.)

derhuerst:
@clee I agree! :D

PR #157, opened 2015-09-25
Added css to showdir.js

avamsi:
Here's what it looks like:

Failed to load!

I am using [http-server](https://github.com/indexzero/http-server) which uses this library.
I wanted to know if you are open to css/icons.

Here's the css file - [style.css](https://gist.github.com/krikx/7ba8164e9faf6d5a963d)


jfhbrook:
idk if you saw but there's [another pull request along similar lines](https://github.com/jfhbrook/node-ecstatic/pull/147) already.

I'll try to review this, but any feedback I gave to #147 likely applies here.

lib/ecstatic/showdir.js
```diff
@@ -110,6 +110,7 @@ module.exports = function (opts, stat) {
...
+            '    <style>@import url(https://fonts.googleapis.com/css?family=Open+Sans); * { margin:0; padding:0; -webkit-box-sizing:border-box; -moz-box-sizing:border-box; box-sizing:border-box; } html { min-height:100%; border:10px solid #ECEEF1; color:#61666c; font-weight:400; font-size:1em; font-family:"Open Sans", sans-serif; line-height:2em; } body { margin:0 auto; max-width:80%; padding:20px; -webkit-backface-visibility:hidden; } h1 { color:#2a2a2a; font-weight:400; margin-bottom:1em; } a { color:#61666c; text-decoration:none; } a:hover { color:#2a2a2a; } table { width:100%; table-layout:fixed; border-collapse:collapse; font-size:.925em; } tr:hover td { background:#f6f6f6; } td { padding:5px 0; outline:0; border:0; overflow:hidden; border-bottom:1px solid #edf1f5; vertical-align:middle; text-align:left; -webkit-transition:background 196ms ease; -moz-transition:background 196ms ease; -ms-transition:background 196ms ease; -o-transition:background 196ms ease; transition:background 196ms ease; } td a { display:block; } address { font-size:80%; text-align:left; font-style:normal; } address a { color:#f32236; text-decoration:none; } address a:hover { color:#f37c4c; }</style>',
```
jfhbrook: Not huge on relying on external resources (ie fonts.googleapis.com) to make this work.

lib/ecstatic/showdir.js
```diff
@@ -129,12 +130,15 @@ module.exports = function (opts, stat) {
             }
 
             var displayName = he.encode(file[0]) + ((isDir)? '/':'');
+            var ifile = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAAN1wAADdcBQiibeAAAABl0RVh0U29mdHdhcmUAd3d3Lmlua3NjYXBlLm9yZ5vuPBoAAAC0SURBVDiN7Y8tjsMwGESn1kgBAQEGoSELs7fojdpb9FShYd4gh4TmCt+PXVReV4U70sD3NHNZ1/VqZo9Syi8aEkJIJG+XZVnSNE3zOI4IIbwFl1JwnieO4/ijiMwxRohIywDEGJFznunucPcmGABeHM0MtdZmAQCYGaiqHwtUFXT3jwXfufAv+JJAVUGyGTYzBJI5pQQRQa31rYoIUkogmdn3/X3f98e2bT8tC7qu24dhuD8BD6e7SzzK9MwAAAAASUVORK5CYII=";
```
jfhbrook: What's the licensing on these images? Where did they come from? Etc?

krikx: @jfhbrook icons used are [Faenza Icons](http://tiheum.deviantart.com/art/Faenza-Icons-173323228) and they are under [GNU General Public License](https://en.wikipedia.org/wiki/GNU_General_Public_License).

jfhbrook: Can you tell me more about how you generated these strings? I might want to use this technique to, say, add more icon types.

krikx: @jfhbrook you just have to encode the image in [Base64](https://en.wikipedia.org/wiki/Base64).
Here's the code I am using:
```python
with open(path_to_image, 'rb') as f:
    print f.read().encode('base64').replace('\n', '')

```

fenwick67:
Hey @krikx I'm the author of #147, I had a similar goal in mind of adding icons and styles, I was just cracking my fingers to work on mine today.

I think maybe somewhere in-between our PRs might be  the best solution.  How this this idea sound? @jfhbrook 

* Add css and style options:
  * `css`: user provides a URL for a `<link>` tag that we put in the head
  * `style`: user provides a string for an inline `<style>` tag that we put in the head
  * if either of those are on, put `<i>` tags in the list with icon classes like in my PR, to enable custom icon styling, maybe add classes to the other elements as well.
* Add option to use a default inline style (call the option `pretty`?).  
  * @krikx's style looks like it would be nice for this because it's relatively small and pretty good looking IMO... but we would change it to use the icon classes mentioned above, and get rid of the `img` elements (we can inline base64 data into the style).

This way we enable easy styling so a user could just just `node-ecstatic --css "/styles/mystyle.css"` or `node-ecstatic --style "body{background:azure}"`

It also avoids the weird folder mirroring I did to handle serving images.

lib/ecstatic/showdir.js
fenwick67: `font-family: "Open Sans",Verdana,Helvetica,sans-serif;` or something similar should get you close enough without having the dependency.

jfhbrook:
I'm totally okay with adding inline css and inline base64-encoded icons as a non-configurable default.

Flags for specifying custom css and/or styles could be interesting, but perhaps better in a separate PR.

jfhbrook:
My thoughts were to take the base64 icon approach from this PR, write a node script for "building" the base65 blobs, apply it to the icon set used in #147 with a build script , choose a set of CSS to inline (probably this one since I'm lazy and it's already inlined), pick up PROTIPS from the other CSS as best as I can, potentially apply styles to the html introduced in #151 and call it a day.

jfhbrook:
I like that #147 doesn't reorder the columns.

krikx:
@jfhbrook I reordered the columns because I don't have much use for permissions.

Also, can you do something like this?
Failed to load!!
Add links in h1 so that we can directly jump to specific folders.
I changed locally but there are other changes in [showdir.js](https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic/showdir.js) like using `he.encode` of which I have no idea about.
(I am using old version)

krikx:
As @fenwick67 suggested, I added those 2 images to `<style>` tag. Source looks much nicer now.
This way many icons can easily be added.

jfhbrook:
Breadcrumbs aren't the worst idea but should be tabled into a separate PR.

jfhbrook:
Closing in favor of #188 . Your name will still be in the contributors file, and hey, feedback appreciated!

PR #158, opened 2015-09-30
opts.headers

substack:
This patch implements curl-like `-H`/`--header` like functionality using `opts.H`/`opts.header`/`opts.headers` for parity between the command-line tool and the API.

This patch replaces #153.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -26,6 +26,7 @@ var ecstatic = module.exports = function (dir, options) {
       baseDir = opts.baseDir,
       defaultExt = opts.defaultExt,
       handleError = opts.handleError,
+      headers = opts.headers,
       serverHeader = opts.serverHeader,
       weakEtags = opts.weakEtags;
 
@@ -73,6 +74,9 @@ var ecstatic = module.exports = function (dir, options) {
       // Set common headers.
       res.setHeader('server', 'ecstatic-'+version);
     }
+    Object.keys(headers).forEach(function (key) {
+      res.setHeader(key, headers[key])
+    })
 
     // TODO: This check is broken, which causes the 403 on the
     // expected 404.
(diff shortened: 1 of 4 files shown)


lib/ecstatic/opts.js
```diff
@@ -90,6 +91,29 @@ module.exports = function (opts) {
...
+        Object.keys(opts[k]).forEach(function (key) {
+          headers[key] = opts[k][key];
+        });
+      }
+      else setHeader(opts[k]);
+
+      function setHeader (str) {
+        var m = /^(.+?)\s*:\s*(.*)$/.exec(str)
+        if (!m) headers[str] = true
+        else headers[m[1]] = m[2]
```
jfhbrook: Did you stop using semis? Honest question. I don't mind cleaning this up.

lib/ecstatic.js
```diff
@@ -80,6 +84,9 @@ var ecstatic = module.exports = function (dir, options) {
       return status[403](res, next);
     }
 
+    if (req.method === 'OPTIONS') {
+      return res.end()
+    }
```
jfhbrook: Technically a breaking change. I wonder who could get hurt by this?

jfhbrook: Might have to feature flag this. Wondering if it's worth doing now or later (I don't mind a major semver bump)

lib/ecstatic/opts.js
```diff
@@ -90,6 +91,29 @@ module.exports = function (opts) {
     });
 
     [
+      'H',
+      'header',
+      'headers'
+    ].forEach(function (k) {
+      if (!isDeclared(k)) return;
+      if (Array.isArray(opts[k])) {
+        opts[k].forEach(setHeader);
```
jfhbrook: Interesting way to do this. I was expecting the array/string part to be special cased to the bin handler.

Not a bad thing.

lib/ecstatic.js
jfhbrook: As an aside, a --cors setting might be cool too. I should probably whip something up post-merge.

jfhbrook:
See: https://github.com/jfhbrook/node-ecstatic/pull/159/files

jfhbrook:
Merged with fixes https://github.com/jfhbrook/node-ecstatic/pull/159 and released in v1.1.0.

Thanks!

PR #160, opened 2015-10-02
boolean options

substack:
Currently:

```
$ ecstatic --cors public -p 8000
```

will serve `./`, NOT `./public` because `--cors` gobbles up the `public` argument. With this patch, the above works as expected.


jfhbrook:
Released in v1.1.1.

Thanks!

PR #161, opened 2015-10-03
more boolean handling

substack:
On the back of #160, this patch fixes an issue where setting an option as boolean sets the value to `false` unless otherwise specified in minimist. I've factored the defaults from opts.js into defaults.json so they can be shared from opts.js and the bin script and also the aliases. Before the handleOptionsMethod was set to `false` even when `--cors` was turned on because of the boolean setting.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -318,11 +318,11 @@ function decodePathname(pathname) {
 }
 
 if (!module.parent) {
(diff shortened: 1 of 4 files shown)


jfhbrook:
Released as v1.1.2.

Thanks! This refactor has been a long time comin'.

PR #162, opened 2015-10-04
set cors to false to enable booleanism on the command-line

substack:
I missed `cors` in the default list, which was the whole reason behind the first patch!

With this patch:

```
$ ecstatic --cors public/ -p 8000
```

will finally work.


jfhbrook:
Published as 1.1.3

Thanks!

PR #166, opened 2015-11-10
Add opts.showDotfiles

camwiegert:
Thought it'd be useful to be able to exclude dotfiles from directory listings. This came up when I wanted to use ecstatic to create a quick local server to share a directory of files with a non-technical coworker. It doesn't affect `../`links.

### DIFF
--- a/lib/ecstatic/showdir.js
+++ b/lib/ecstatic/showdir.js
@@ -13,6 +13,7 @@ module.exports = function (opts, stat) {
       baseDir = opts.baseDir,
       humanReadable = opts.humanReadable,
       handleError = opts.handleError,
+      showDotfiles = opts.showDotfiles,
       si = opts.si,
       weakEtags = opts.weakEtags;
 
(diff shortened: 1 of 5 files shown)


lib/ecstatic/showdir.js
```diff
@@ -40,6 +41,14 @@ module.exports = function (opts, stat) {
         if (err) {
           return handleError ? status[500](res, next, { error: err }) : next();
         }
+
+        // Optionally exclude dotfiles from directory listing.
+        if (true) {
```
jfhbrook: this doesn't seem so optional. ;D

camwiegert: Well, that's embarrassing. Sorry, that was for ad hoc testing!

camwiegert:
Should be slightly more optional now. :)

jfhbrook:
Cool, thanks

jfhbrook:
Released in 1.3.0

PR #167, opened 2015-11-14
Add showDotfiles to main opts example in README.

camwiegert:
Adds showDotfiles to main opts example in README. Also took the opportunity to update the CONTRIBUTORS.

```js
var opts = {
             root               : __dirname + '/public',
             baseDir            : '/',
             cache              : 3600,
             showDir            : true,
             showDotfiles       : true,
             autoIndex          : false,
             humanReadable      : true,
             headers            : {},
             si                 : false,
             defaultExt         : 'html',
             gzip               : false,
             serverHeader       : true,
             contentType        : 'application/octet-stream',
             mimeTypes          : undefined,
             handleOptionsMethod: false
           }
```


jfhbrook:
Thanks! I totally forgot about the CONTRIBUTORS file (it's pretty new)

PR #169, opened 2015-12-10
Allow setting mimeTypes via cli

joshgillies:
This patch enables the following, for cases where maintaining a .types file is overkill:

`ecstatic ./data --port 1337 --mimeTypes '{ "application/x-my-type": ["x-mt"] }'`

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -36,6 +36,9 @@ var ecstatic = module.exports = function (dir, options) {
 
   // Support hashes and .types files in mimeTypes @since 0.8
   if (opts.mimeTypes) {
+    try {
+      opts.mimeTypes = JSON.parse(opts.mimeTypes);
+    } catch (e) {}
     if (typeof opts.mimeTypes === 'string') {
       mime.load(opts.mimeTypes);
     }
--- a/test/custom-content-type-json-string.js
+++ b/test/custom-content-type-json-string.js
@@ -1,0 +1,30 @@
+var test = require('tap').test,
+    http = require('http'),
+    request = require('request'),
+    ecstatic = require('../');
+
+test('custom contentType as stringified JSON. Ref: #169', function(t) {
+  try {
+    var server = http.createServer(ecstatic({
+      root: __dirname + '/public/',
+      mimetype: JSON.stringify({
+        'application/jon': ['opml']
+      })
+    }));
+  } catch (e) {
+    t.fail(e.message);
+    t.end();
+  }
+
+  t.plan(3);
+
+  server.listen(0, function() {
+    var port = server.address().port;
+    request.get('http://localhost:' + port + '/custom_mime_type.opml', function(err, res, body) {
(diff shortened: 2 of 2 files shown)


dotnetCarpenter:
I would like verify this before merge. It looks good but doing a local test would make me more confident about this PR.

joshgillies:
Hey @dotnetCarpenter, I've just pushed a test. Please, let me know if there's anything else I should do here.

dotnetCarpenter:
Hi @joshgillies . I'm currently stuck on issue #172. But will look through the files here on github. It occurs to me that you're testing the mime type option in ecstatic setup which has always worked and is tested here: https://github.com/jfhbrook/node-ecstatic/blob/master/test/custom-content-type.js

But your PR is about the cli which isn't tested. I can help you writing a test for it as soon as #172 is fixed. You need to use `require('child_process').spawn` or similar.

dotnetCarpenter:
Ahh, I see you added to the option as JSON. I'm afraid that is a bad test.  We want to test exactly the feature you're implementing not a backdoor that runs the same call path.

joshgillies:
OK, I'll add a cli specific test for this. :smile:

dotnetCarpenter:
@joshgillies Do  you need any help with writing the test?

jfhbrook:
Keep in mind that afaik there are no CLI tests whatsoever.

Adding them probably doesn't have to be a full-on integration test situation, we should be able to parse the cli flags using our correctly-factored functions and then just assert.

If you want to just write the one test, that's fine, but don't let me stop you from writing something a little more comprehensive ;D

dotnetCarpenter:
@jfhbrook I'm just thinking that if we want to refactor the cli part, then the only useful test we'll have will be the ones that actually test the cli part from outside and not the ones that test the functions from inside, which at that point probably need to be rewritten and there goes the confidence in not creating any regression.

I think it is super easy to do but I might be wrong. Let take a stab at it and see.

jfhbrook:
What I have in mind is factoring https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L332-L341 into an exported function and assuming that these opts get passed to the middleware correctly.

joshgillies:
Refactoring the CLI component out into it's own component sounds good to me. @dotnetCarpenter I didn't have a huge amount of spare time over the weekend and didn't get around to adding an appropriate test for this. If you'd like to take a stab, feel free. :smile:

dotnetCarpenter:
@joshgillies I've merged in your PR into mine and wrote a CLI test for your changes. I also wrote one for adding a file but I actually think now, that it is redundant. I'll have to check but pretty sure we already got a test for that. I wrote two tests for changing ports but for some reason they fail on travis but not locally. So I turned them off. The main thing is that there is a CLI test for your changes and it passes.

In other words, this PR looks good to me :+1: 
@jfhbrook Feel free to merge in this or #178

PR #173, opened 2015-12-11
Added documentation about cloning ecstatic on Windows - close #172

### LINKED ISSUE #172: The <dir> folder in test/public is breaking git in Windows
One can not have a folder named <dir> in Windows, which prohibit me from cloning and updating my local copy of `node-ecstatic`. Can we called it something else, please?

On Windows the following characters are forbidden in a folder name; `\ / < > | * ? " :`

I currently can not checkout pull requests.

dotnetCarpenter:
I think the simplest way to overcome our tests, for malformed directory listings links, breaking git on Windows, is to educate Windows users about the caveat. This is my attempt to do so.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,49 @@ discretion.
 
 Glad we cleared that up.
 
+## Windows users
+
+Before you clone ecstatic you unfortunately have to make some changes to git.
+In our test suit we test for certain named folders, that need proper
+encoding to not break links in directory listings. These folder names are
+illegal on Windows and will throw off git. It's a catch-22 because you need
+to make these changes to your local ecstatic git repository, which you can
+not clone.
+
+1) Create and initialize your new repository (`<url>` is your fork):
+
+```
+mkdir node-ecstatic
+cd node-ecstatic
+git init
+git remote add –f origin <url>
+```
+
+2) Enable sparse-checkout:
+
+```
+git config core.sparsecheckout true
+```
+
+3) Configure sparse-checkout by listing your desired and excluded sub-trees
+   in .git/info/sparse-checkout:
+
+```
+echo /*
+echo !test/public/<dir> >> .git/info/sparse-checkout
+echo !test/showdir-search-encoding.js >> .git/info/sparse-checkout
+echo !test/showdir-pathname-encoding.js >> .git\info\sparse-checkout
+```
+
+4) Checkout from the remote:
+
+```
+git pull origin master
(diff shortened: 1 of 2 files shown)


dotnetCarpenter:
@jfhbrook Is my english that bad?

jfhbrook:
Nah I just let this get lost in the shuffle.

CONTRIBUTING.md
```diff
@@ -16,6 +16,49 @@ discretion.
...
+```
+
+3) Configure sparse-checkout by listing your desired and excluded sub-trees
+   in .git/info/sparse-checkout:
+
+```
+echo /*
+echo !test/public/<dir> >> .git/info/sparse-checkout
+echo !test/showdir-search-encoding.js >> .git/info/sparse-checkout
+echo !test/showdir-pathname-encoding.js >> .git\info\sparse-checkout
```
jfhbrook: Is this in bash or powershell? What would this look like if I just wanted to crack open notepad?

jfhbrook: If it's bash, it doesn't seem to work for me. Maybe quotes?

jfhbrook: Okay, I got this to work with vim, I see what's up.

jfhbrook:
Are there characters which would be good candidates for checking encoding that are in fact legal in windows and mac/linux ?

dotnetCarpenter:
Yes, the Windows illegal characters are **\ / < > | * ? " :**

I also posted them in #172

dotnetCarpenter:
Sorry, misread your question

dotnetCarpenter:
@jfhbrook as I wrote in #172, the attack we're protecting us against is only made possible by using characters that are not legal on Windows. So in order to check that ecstatic is in fact encoding these directory names, you have to have a directory that is illegal to have on Windows. I can't think of a way out. Perhaps someone knows in some security forum?

CONTRIBUTING.md
dotnetCarpenter: It's not bash it's Windows cmd. :scream:

jfhbrook: Oh dang! Oh like the command line, I didn't know it supported >>

jfhbrook: Luckily even with my ill-informed edits the directions should still work. *whew*

jfhbrook:
I *think* the characters just have to be ones that need to be html encoded, yeah?

At any rate I agree, this should be documented.

PR #182, opened 2015-12-30
make code coverage work on windows

isaacs:
Fixes https://github.com/isaacs/node-tap/issues/199


jfhbrook:
Does this also fix the "can't clone in windows" problem?

jfhbrook:
Also, I use semis.

isaacs:
> Does this also fix the "can't clone in windows" problem?

It seems to, yes!  That `<dir>` directory, windows does not like it.

I'll add semis and re-push.

jfhbrook:
Interesting. Right now CONTRIBUTING has directions on doing a sparse checkout to avoid that problem, but creating the file on-the-fly obviously works too.

isaacs:
It also means skipping the test, because Windows does not like it.

```
> fs.mkdirSync('<dir>')
Error: ENOENT: no such file or directory, mkdir 'c:\ecstatic\<dir>'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:794:18)
    at repl:1:4
    at REPLServer.defaultEval (repl.js:248:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:412:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
```

isaacs:
Ok, travis likes it.  LGTM if you're ok with it.

dotnetCarpenter:
@isaacs Maybe I'm being darf but all I get when doing `npm test` now is `Reading TAP data from stdin (use "-" argument to suppress)`. Is this the expected behavior?

dotnetCarpenter:
Hmm, I guess I am. My test folder is empty. It's probably a local issue

PR #185, opened 2016-01-08
Do not redirect / add trailing slash when showDir and autoIndex are off.

sstur:
When both `showDir` and `autoIndex` options are OFF we do not serve directories in any form. So if the client requests `/foo` and `foo` is a directory, we should NOT handle the request at all. It doesn't make sense to redirect them to `/foo/` because we won't be able handle that path either. This creates an un-necessary round-trip and perhaps more importantly it exposes details about our filesystem structure.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -124,6 +124,11 @@ var ecstatic = module.exports = function (dir, options) {
           status[500](res, next, { error: err });
         }
         else if (stat.isDirectory()) {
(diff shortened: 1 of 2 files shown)


dotnetCarpenter:
looks good

jfhbrook:
Sorry, just now seeing this.

Thanks!

sstur:
Cool. Thanks!

PR #186, opened 2016-02-10
utf8 forbidden in headers since nodejs 4.3.0

igorklopov:
national characters are used in date if non-US locale
https://github.com/nodejs/node/blob/v4.3.0/lib/_http_common.js#L219


jfhbrook:
That's annoying, but okay.

Can you add a test and your name to the CONTRIBUTORS.md document?

jfhbrook:
If testing is Hard don't sweat it.

igorklopov:
This is a regression and it will be fixed. https://github.com/nodejs/node/pull/5237

PR #187, opened 2016-03-12
Improve correctness and efficiency of cache validation code

jessetane:
I recently noticed that I was only ever getting 200's from ecstatic and tracked the issue down to `stat.mtime` having millisecond resolution on my system, which is more than what's included in the date string sent in the last-modified header.

The test doesn't fail because it's not doing what I think browsers do and what the spec seems to be recommending:
> To get best results when sending an If-Modified-Since header field for cache validation, clients are advised to use the exact date string received in a previous Last-Modified header field whenever possible.

I updated the test accordingly and discovered it still didn't fail on my laptop, where mtimes seem to be lower resolution, but maybe it will fail at travis?

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -227,7 +227,8 @@ var ecstatic = module.exports = function (dir, options) {
 
       // TODO: Helper for this, with default headers.
       res.setHeader('etag', etag(stat, weakEtags));
-      res.setHeader('last-modified', (new Date(stat.mtime)).toUTCString());
+      var lastModified = (new Date(stat.mtime)).toUTCString();
+      res.setHeader('last-modified', lastModified);
 
       if (typeof cache === 'function') {
         var requestSpecificCache = cache(pathname);
@@ -240,7 +241,7 @@ var ecstatic = module.exports = function (dir, options) {
       }
 
       // Return a 304 if necessary
-      if (shouldReturn304(req, stat)) {
+      if (shouldReturn304(req, stat, lastModified)) {
         return status[304](res, next);
       }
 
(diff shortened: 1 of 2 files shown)


jessetane:
Ok cool, looks like it fails on linux, I'll propose a fix.

jessetane:
Ok should pass now, not the most elegant thing but should be accurate and avoids calling `toUTCString()` twice?

jfhbrook:
It's been a while since I've dug into this so I'm a little confused. Who sends the last-modified header, the client? *shouldn't* we be comparing against file mtime?

jfhbrook:
Oh wait, *we* set this on L230? Right?

jessetane:
Yeah, so on L230 the server sets the last-modified header, dropping resolution as it goes to UTC string, which I'm pretty sure is correct and has to happen every time. The change I made on L295 then converts that string back to a Date for foolproof comparison with the client's if-modified-since, which should generally only ever be a straight copy of a previously seen last-modified.

My fix obviously isn't so legible - it would be clearer and faster if we had a variable around to hold the mtime string we generated for setting last-modified, but I wasn't sure where to put it. The getHeader() business is still faster than another call toUTCString() so I just went with that.

jessetane:
If you have advice on where the variable should go, happy to send another patch.

jfhbrook:
Probably at the top of the narrowest scope where it can still be used in all the proper places. So maybe near where I set the header?

jessetane:
hah, I think the code is better, but the commit is probably more confusing :|

jessetane:
hang on I'll split this up

jessetane:
ok maybe this

jessetane:
To accommodate for the scope creep I changed the issue title. Hopefully it still makes sense to have this all in one PR.

jfhbrook:
@jessetane released in 1.4.1. Sorry for the mad delay!

Thanks!

jessetane:
Nice!

PR #188, opened 2016-08-10
CSS icons

jfhbrook:
Borrowing a little of #157 (the inline base64 icons) and of #147 (base css params and similar approach otherwise), with a [different set of CSS icons](https://github.com/teambox/Free-file-icons/tree/master/32px).


jfhbrook:
The different set of CSS icons is MIT licensed, by the way, so we should be good there.

PR #146, opened 2015-08-12
Add `fallbackFile` option to response for unhandled requests

ngs:

PR #190, opened 2016-09-06
Use 200.html as a fallback before 404.html

geelen:
This is a convention I think makes a lot of sense for single page applications, when a URL doesn't match anything.

- First check for `/200.html` and serve that with status `404`
- Then check for `/404.html` and serve that with status `404`
- Then serve a blank `404`

This was introduced by [harp](https://github.com/sintaxi/harp) (I think) and is used by [surge.sh](https://surge.sh/help/adding-a-200-page-for-client-side-routing) and it seems to be really straightforward to use.

I've made it so that if you set the config of `handleError` to `false` then this doesn't look for `200.html` either, just to keep compatibility with other clients.


jfhbrook:
Hey, just now seeing this. Will try to review in the next few days.

lib/ecstatic.js
```diff
@@ -124,13 +124,26 @@ var ecstatic = module.exports = function (dir, options) {
...
             middleware({
-              url: (handleError ? ('/' + path.join(baseDir, '404.' + defaultExt)) : req.url),
+              url: '/' + path.join(baseDir, '404.' + defaultExt),
               statusCode: 404
             }, res, next);
           }
+          else {
+            // Try ./200.html before falling back to ./404.html
+            middleware({
+              url: '/' + path.join(baseDir, '200.' + defaultExt),
```
jfhbrook: This comma looks extraneous.

lib/ecstatic.js
```diff
@@ -124,13 +124,26 @@ var ecstatic = module.exports = function (dir, options) {
...
-          else {
-            // Try to serve default ./404.html
+          else if (!handleError) {
+            // If we're not handling errors/fallbacks, bail out now
+            middleware({
+              url: req.url,
+              statusCode: 404
+            }, res, next);
+          }
+          else if (path.basename(parsed.pathname, defaultExt) === '200.') {
```
jfhbrook: This check seems brittle and awkward, and I don't think anyone else does it. Do you see a way around it? I don't, least not without a bigger refactor of ecstatic around passing state (which I want to do, but is kind of a whole thing).

I'll have to think about this.

jfhbrook: Also: Can we put this behind a flag? I don't know what to call it, but this should be opt-in behavior, at least for now.

lib/ecstatic.js
```diff
@@ -124,13 +124,26 @@ var ecstatic = module.exports = function (dir, options) {
...
+            // Already tried ./200.html, now try ./404.html
             middleware({
-              url: (handleError ? ('/' + path.join(baseDir, '404.' + defaultExt)) : req.url),
+              url: '/' + path.join(baseDir, '404.' + defaultExt),
               statusCode: 404
             }, res, next);
           }
+          else {
+            // Try ./200.html before falling back to ./404.html
```
jfhbrook: It's mildly surprising that this isn't opt-in behavior, but 200.html wasn't used for anything else prior, right? So this is a breaking change, but that seems fine.

jfhbrook:
So, uh, yeah! Github's new review feature is *okay*.

lib/ecstatic.js
```diff
@@ -124,13 +124,26 @@ var ecstatic = module.exports = function (dir, options) {
               url: parsed.pathname + '.' + defaultExt + ((parsed.search)? parsed.search:'')
             }, res, next);
           }
-          else {
-            // Try to serve default ./404.html
+          else if (!handleError) {
+            // If we're not handling errors/fallbacks, bail out now
```
jfhbrook: ...is this a bug? I'll have to look more closely, but I think maybe handleError: false should have us be calling next directly.

jfhbrook:
Last thought: Is this testable?

jfhbrook:
I lied:

Do you think this PR solves a similar problem as https://github.com/jfhbrook/node-ecstatic/pull/146 ?

slorber:
@jfhbrook it solves similar problem for me in a less flexible way.

In my case, I'd love to be able to give custom fallback page, and not rely on convention 200/404.html pages

mk-pmb:
I wouldn't have expected any special 200/404 file behaviour either, as the README doesn't mention it.

jfhbrook:
Feel free to PR some clarifying docs @mk-pmb

PR #195, opened 2017-05-23
*.html should support gzip

wood1986:
if gzip flag is turned on, it should apply to all assets including *.html. Does it make sense?
if you agree, i will add the test


jfhbrook:
Yeah, absolutely--if certain assets aren't getting gzipped when it's on, then that's a bug.

jfhbrook:
Though: Having gzip on doesn't necessarily mean that the extension is changed to .gz...this is for reading gzipped files off disk? If so, that makes sense.

jfhbrook:
otoh, you could just set the default extension to .html.gz

hmm

wood1986:
But your implementation is based on checking the ".gz" at line 207 to tell client whether it is gzipped content,

Here is my thought. 
if gzip is on, try to get the gzip version first (let's say `ecstatics.html.gz`)
else fallback to `ecstatics.html`. it should apply to all assets.

In short, what do you think if I add the fallback logic?

jfhbrook:
you're right, the existing behavior is to take the full filename and add .gz to the end of it when searching for the fs.

Given that, I think the correct behavior is exactly the same as with non-defaulted-extensions, except with '.html' (or whatever) shoved in there. I think this is what you originally had implemented, but I'm at work so can't double-check rn.

dotnetCarpenter:
I don't want to put too much work on your plate. But according to [RFC 1952](http://www.gzip.org/zlib/rfc-gzip.html#header-trailer), a GZIP file always starts with the same 2 bytes to identify it as GZIP: ID1 = 31 (0x**1f**, \037), ID2 = 139 (0x**8b**, \213). - I've highlighted what you will see as the first 2 bytes in nodejs when reading a Buffer.

```js
fs.readFile('./package.json.gz', (status, fd) => console.log(fd)) // -> <Buffer 1f 8b ...
```

I think a proper solution is to use something similar to the above when reading a file (regardless of extension) and if the first 2 bytes are `1f 8b`, then it's gzip and should be handled as such. You could use the gzip flag to opt in for the GZIP sniffing, but I imagine that it's so fast that you'll just want to have ecstatic always being able to tell GZIP from non GZIP - e.i. **fast** since you're streaming the file anyway.

dotnetCarpenter:
Here a gzip file you can use as test data.
[package.json.gz](https://github.com/jfhbrook/node-ecstatic/files/1024131/package.json.gz)

jfhbrook:
Magic bytes is a good idea. We might want to feature flag it to avoid a major version release.

Let me think about this and I'll try to say something intelligent about how to move forward tonight.

wood1986:
my previous claim, "Although I have not tried, I guess *.html.gz cannot be access via *.html" was completely wrong.

If you have a.html and a.html.gz and you hit a.html, ecstatic is able to pick up a.html.gz when gzip flag is on.

Actually the problem I have is on the index.html. By default, ecstatic will pick the index.html when you don't specifiy the file name in the URL. The bug is here, it won't pick the gzip version even if you have the index.html.gz

The line 123, 129 and 159 do not include req.headers. This causes

jfhbrook:
I agree. Making autoindex respect .gz as everything else does is totally a patch I would accept, and we can scope out magic bytes support as a separate issue.

wood1986:
check this out when you are free before I write the test

jfhbrook:
On the face of it this looks ok, but it's hard to tell 100% without seeing tests. :)

wood1986:
it's done

lib/ecstatic.js
```diff
@@ -330,6 +339,19 @@ var ecstatic = module.exports = function (dir, options) {
...
 
+function hasGzipId12(gzipped, cb) {
+  var stream = fs.createReadStream(gzipped, { start: 0, end: 1, autoClose: true }),
+      isGzip = false;
+
+  stream.on("close", function () {
+    cb(isGzip);
+  });
+
+  stream.on("data", function (chunk) {
```
jfhbrook: Possible bug for a stream only giving one byte in the chunk? Unlikely, but I don't think fs makes any guarantees about this.

I've used concat-stream for this kind of thing in the past, it might be helpful for you here.

test/fixtures/common-cases.js
```diff
@@ -68,8 +68,13 @@ module.exports = {
   },
   'subdir/' : {
     code : 200,
-    type : 'text/html',
-    body : 'index!!!\n'
+    headers: { 'accept-encoding': 'compress, gzip' },
```
jfhbrook: Can we make this a separate test, instead of changing an existing one?

test/fixtures/common-cases.js
```diff
@@ -68,8 +68,13 @@ module.exports = {
...
+    body: fs.readFileSync(path.join(__dirname, '../', 'public', 'subdir', 'index.html.gz'), 'utf8')
+  },
+  'subdir/a' : {
+    code : 404,
+    headers: { 'accept-encoding': 'compress, gzip' },
+    body: require('eol').lf(fs.readFileSync(path.join(__dirname, '../', 'public', '404.html.gz'), 'utf8'))
```
jfhbrook: You should require eol at the top of the test file and then `eol.lf` here

test/fixtures/common-cases.js
```diff
@@ -108,6 +113,31 @@ module.exports = {
   'containsSymlink/': {
     code: 404,
     body: '<h1>404</h1>\n'
+  },
+
+  'gzip/real_ecstatic': {
```
jfhbrook: Can we do some asserts for content-type headers?

Note that if any tests that used to pass pre-change and start to fail post-change are potentially breaking, and need to be discussed, so you should confirm how they behave prior to your PR.

lib/ecstatic.js
```diff
@@ -204,7 +213,7 @@ var ecstatic = module.exports = function (dir, options) {
         }
       }
 
-      if (path.extname(file) === '.gz') {
+      if (file === gzipped) {
```
jfhbrook: I know the current codebase isn't a shining example for this kind of thing, but how would you feel about making this a flag instead? `gzipped` being the filename and `isGzipped` being a boolean? I *think* that would be more clear.

lib/ecstatic.js
jfhbrook: You might also want to see if readFile or even readFileSync offer the same API for selecting file ranges.

lib/ecstatic.js
```diff
@@ -85,7 +85,7 @@ var ecstatic = module.exports = function (dir, options) {
             )
           )
         ),
-        gzipped = file + '.gz';
+      gzipped = file + '.gz';
```
jfhbrook: This tabbing looks wrong btw--it should line up with 'file'

lib/ecstatic.js
wood1986: I would prefer a comment over `server(stat, isGzipped)` or `stat.isGzipped = true`

lib/ecstatic.js
wood1986: `fake_ecstatics.gz` is 0 byte file. it works fine.

test/fixtures/common-cases.js
wood1986: resolved

test/fixtures/common-cases.js
wood1986: resolved

test/fixtures/common-cases.js
wood1986: resolved

lib/ecstatic.js
wood1986: resolved

lib/ecstatic.js
jfhbrook: Right, but my concern isn't short files, it's long files.

(I'll preface this by saying that I did some research on whether readFile supports start and end; unfortunately it doesn't. So we can't do that.)

My concern is this: I don't think streams guarantee any minimum length on data buffers, meaning it's possible to get a data event with just the first byte but not the second.

This can be solved with https://www.npmjs.com/package/concat-stream pretty easily, and also help us avoid having to listen for stream events directly which isn't the funnest.

lib/ecstatic.js
jfhbrook: I'd accept a comment.

lib/ecstatic.js
wood1986: How about this?

```
  var stream = fs.createReadStream(gzipped, { start: 0, end: 1, autoClose: true }),
      buffer = Buffer.from([]);

  stream.on("close", function () {
    cb(buffer[0] == 31 && buffer[1] == 139);
  });

  stream.on("data", function (chunk) {
    buffer = Buffer.concat([buffer, chunk], 2);
  });
```

No need to introduce a dependency

wood1986: I don't mind you say no

jfhbrook: looking at concat-stream it looks like its primary differences are around encoding--if we only work with buffers and nothing else, then this is mostly ok.

The one thing you're missing is a stream error handler, and protection against double-callback-ing and other corner cases. I also run travis tests against 0.12, which affects what buffer methods you can use. So something more like,

```js
var stream = fs.createReadStream(gzipped, { start: 0, end: 1, autoClose: true }),
    buffer = Buffer(''),
    hasBeenCalled = false;

stream.on('data', function (chunk) {
  buffer = Buffer.concat([buffer, chunk], 2)
});

stream.on('error', function (err) {
  if (hasBeenCalled) {
    throw err;
  }
  hasBeenCalled = true;
  cb(err);
});

stream.on('close', function () {
  if (hasBeenCalled) {
    return;
  }
  hasBeenCalled = true;
  if (buffer.length !== 2) {
    return cb(null, false);
  }
  return cb(null, buffer[0] == 31 && buffer[1] == 139);
});
```

jfhbrook: We can talk about removing 0.12 support, given that using new Buffer is apparently deprecated because node core likes breaking shit.

wood1986: Ignoring the error is intentional

* node API mentions If autoClose is set to true (default behavior), on error or end the file descriptor will be closed automatically. (I should remove `{ autoClose: true }`). if an `error` is emitted, `close` will be emitted too because of `{ autoClose: true }`

* `buffer[0] == 31 && buffer[1] == 139);` is still safe and no need to the length check even if the buffer is empty. It saves some operations when the file is a real gzip.

* As for the `cb` with `err` argument, if there is an error, I think the cb does not need to include `err` because the result must be `cb(false)`. Meanwhile, `statFile()` will finally handle the error.

* Buffer('') resolves the 0.12 issue.

jfhbrook: As a matter of interface, all callbacks must be of the form `(err, result)` even if that error is always null. I don't really like the idea of swallowing errors, but buy the argument that the right behavior overall is to swallow errors and callback false. However, last I checked an emitted error event on a stream will throw if you don't explicitly add an error handler, so even if you callback null, false you still need to listen for that event.

I confirmed the buffer length check is unnecessary. Weird that buffers allow for that, but ok.

wood1986: Thanks for your time for the discussion and review. I hope this can meet your requirement.

lib/ecstatic.js
```diff
@@ -330,6 +339,24 @@ var ecstatic = module.exports = function (dir, options) {
...
+
+  stream.on("error", function (err) {
+    error = err;
+  });
+
+  stream.on("data", function (chunk) {
+    buffer = Buffer.concat([buffer, chunk], 2);
+  });
+
+  stream.on("close", function () {
```
jfhbrook: Docs say close is not called when error is emitted. You have to cb separately in the error case.

wood1986:
I did fix it. Please review

wood1986:
Thanks for the approval

PR #199, opened 2017-08-20
Use external stylesheets in showdir

mahdi-ninja:
Using external css file in directory listing.

### DIFF
--- a/lib/ecstatic/show-dir/index.js
+++ b/lib/ecstatic/show-dir/index.js
@@ -24,6 +24,15 @@ module.exports = function (opts, stat) {
 
   return function middleware (req, res, next) {
 
+    if (req.url.endsWith('?css')) {
+      res.writeHead(200, {
+        'Content-Type': 'text/css',
+        'Cache-Control': 'public, max-age=31536000'
+      });
+      res.end(css);
+      return;
+    }
+
     // Figure out the path for the file from the given url
     var parsed = url.parse(req.url),
         pathname = decodeURIComponent(parsed.pathname),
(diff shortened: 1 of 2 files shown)


jfhbrook:
Interesting.

I was actually going to approach this by saying that, no, the CSS should be inline, so that there isn't a separate route/request. I personally have zero issues leaving that as-is, and I kinda wish I deleted this comment.

But this appears to switch on a querystring param, which kinda solves that problem but also introduces, well, I've never seen content-type switch on a qs param like this. I feel like it's weird.

How important do you think this kinda change is? Like did you see the comment and think "well fuck I can do that", or do you have an ideological horse in this race? I'd just as soon leave this one out, but willing to have the discussion.

Like with #200, also broke this by merging the airbnb branch, sorry

lib/ecstatic/show-dir/index.js
```diff
@@ -24,6 +24,15 @@ module.exports = function (opts, stat) {
 
   return function middleware (req, res, next) {
 
+    if (req.url.endsWith('?css')) {
+      res.writeHead(200, {
+        'Content-Type': 'text/css',
+        'Cache-Control': 'public, max-age=31536000'
```
jfhbrook: I feel like ideally cache headers should behave the same as in the rest of the app.

I get that maybe directory listings don't properly do this now, but that would also be a bug.

mahdi-ninja: I though because the styles are built into the server and are not going to change anytime soon, it makes sense to cache them for a long time.

mahdi-ninja:
Of course this is not that important. I just think it's ugly to include styles in every response. TBH, I saw your comment after I started editing it ;)

About using query string for serving different content types, you're right. It's not common. But I don't see any reason why we can't use it this way. According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4):
>The query component contains non-hierarchical data that, along with data in the path component, serves to identify a resource within the scope of the URI's scheme and naming authority.

mahdi-ninja:
By the way, let me know if you wanted to merge this and I'll rebase it on master to fix the conflicts and have a clean history.

mk-pmb:
> About using query string for serving different content types, you're right. It's not common.

I think it's rather widespread for APIs do deliver either JSON or JS depending on whether the query params specify a JSONP callback function name.

jfhbrook:
Closing this for cleanup reasons, since I don't think this will get merged. Thanks tho!

PR #200, opened 2017-08-20
Fix icon styles in showdir when text is wrapping in multiple lines

mahdi-ninja:
Before fix:
Before fix

After fix:
After fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Remove support for 0.12.0
 - Remove union examples and test harnesses (support should have been removed
   long ago)
+- Fix icon styles in directory listing for small screens
 
 2017/06/06 Version 2.2.1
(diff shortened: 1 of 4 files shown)


jfhbrook:
This looks nice.

Can you do me a solid and update the changelog and contributors list before I merge?

jfhbrook:
You unfortunately (well, fortunately moving forward, but for the purposes of this pr...) reminded me to merge a "we use modified airbnb now" PR, so there will have to be some manual work to make this merge cleanly.

If you can make it work, great, but if not, I can do the work, so don't stress too hard about the grey stuff if you can't fix it.

mahdi-ninja:
I rebased my changes on master and updated changelog and contributors list.

jfhbrook:
Thanks!

PR #203, opened 2017-10-21
Allow to serve static files as response to POST

yura-pakhuchiy:
Sometimes it is useful to serve static answers to POST requests. However ecstatic rejects all requests except GET or HEAD. This small patch fixes it.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -221,7 +221,7 @@ module.exports = function createMiddleware(_dir, _options) {
       return;
     }
 
-    if (req.method && (req.method !== 'GET' && req.method !== 'HEAD')) {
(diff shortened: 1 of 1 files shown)


jfhbrook:
I, uh, why would you want this? POSTs should *do* something. At the least, this behavior seems technically incorrect.

yura-pakhuchiy:
To test application which uses HTTP API. I can create directory which contains replies to requests and point application to it. Without this patch I can test only GET requests, however POST request are very common as well.

While I agree that requests that can change something should use POST, there is no requirement for every POST request to change something. Consider POST request with wrong access token, server should just ignore them.

BTW, PHP  web server (`php -S`) supports both GET and POST for serving files. However I think it is bit overkill to install PHP just to use `php -S`, that is why I want similar functionality in node based project.

PR #207, opened 2017-12-09
fixed CSS and layout

DRSDavidSoft:
1. Using `background` instead of `background-image` will override previously set attributes such as `background-repeat: no-repeat` which is undesirable

2. Since the table is not automatically width-fixed, we have to disable line breaking in order to show the entire path (and permissions) in one line

## Preview
fix-1


jfhbrook:
Looks reasonable. Thanks! Can you do me a favor and add yourself to the CONTRIBUTORS file? Otherwise I can do it.

DRSDavidSoft:
Sure, gladly. I added it to my commits.

benevbright:
By the way, is there any option to hide permissions of files?
Thanks.

Issue #208, opened 2017-12-16
Add a logging option


dotnetCarpenter:
I don't know if I'm the only one who needs this, but because ecstatic is my main development server, I often wish I could see the requests in the console. I would like to be able to enable it with a simple CLI flag.

For log format I was thinking something like https://en.wikipedia.org/wiki/Common_Log_Format

jfhbrook:
Yeah I could get behind a simple CLF-based logging format

mk-pmb:
How about we integrate [morgan](https://www.npmjs.com/package/morgan)? It offers a variety of log formats.

dotnetCarpenter:
@mk-pmb interesting. I'll take a look

mk-pmb:
[Example code](https://gist.github.com/mk-pmb/e982697487de3f501705b73b579dd72f) for ecstatic `stack`ed with logging.

jfhbrook:
I think it's a reasonable stance to say that ecstatic's logging should be extremely lightweight and intended for stdout only. If this means pulling in a module this is OK, but should definitely not overengineer this. If a tiny custom class works just as well, we should consider that as well.

mk-pmb:
Alternatively, we could consider implementing a very light-weight sub-middleware chaining mechanism inside the middleware function we return, so people can plug whatever logging and more. ([Yo dawg](http://knowyourmeme.com/memes/xzibit-yo-dawg), I heard you like middleware…)

jfhbrook:
idk about sub-middlewares (I've been toying with the idea of moving away from express middleware support for well over a year now anyway), but a `log` property on the existing middleware to hook onto would make a lot of sense.

Issue #193, opened 2017-04-11
Unused code in handling 405


casr:
I was reading the source code surrounding a 405 response and it stuck me that there is an undocumented option, `opts.allow`<sup>[1]</sup>, that would also be impossible to fire at this time<sup>[2]</sup> as no options are passed to the 405 error handler:

[1]: https://github.com/jfhbrook/node-ecstatic/blob/4ba99b7/lib/ecstatic/status-handlers.js#L30
[2]: https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L108-L110 

```js
// disallowed method
exports['405'] = function (res, next, opts) {
  res.statusCode = 405;
  if (typeof next === "function") {
    next();
  }
  else {
    res.setHeader('allow', (opts && opts.allow) || 'GET, HEAD');
    res.end();
  }
};
```

```js
    if (req.method && (req.method !== 'GET' && req.method !== 'HEAD' )) {
      return status[405](res, next);
    }
```

Issue #154, opened 2015-09-18
Option to yield to next middleware without setting response code 


ryanstevens:
This is more of a question at this point, but why does ecstatic set the response code in some cases before it yields control to the next middleware in the chain.  So for example if estatic hits the 404 use case this [status-handler#L37](https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic/status-handlers.js#L37) sets the response code to 404 AND then calls next() if it exists.  

I would prefer to pass in an option to not set the response code if another middleware will ultimately be invoked.  Philosophically I would make the case if estatic yields control to another middleware it shouldn't mutate anything on the response object since it "had it's chance" so to speak.

jfhbrook:
There's two things going on here. One is that this is a way for downstream middlewares to detect what happened inside ecstatic---without setting the status code there isn't really a good way to know *why* a request fell through. The other is an artifact of architecture---ecstatic has a few recursive calls, and I *believe* (it's been a while since I've looked at this) it branches based on req state.

Unfortunately, ecstatic is constrained by the fact that it's intended to be used as a middleware. That has unfortunate ramifications both pragmatic and philosophical beyond but including setting of status codes as a mechanism for custom 404/etc handling.

Issue #88, opened 2013-10-14
Cannot generate file listing if `404.html` is present in /


leesei:
Found this issue when running `ecstatic` in `test/public`.
Response is served when the middleware in `autoIndex` switch found `404.html` while trying to serve `index.html`, so `showDir()` is never called.

Because of this issue, I cannot add test case for `showDir()` (so #86 was not covered in the test).

PR #209, opened 2017-12-16
[WIP] Logging

dotnetCarpenter:
This is my initial idea on how to provide logging for ecstatic.

The PR is only half done but the tests and main output is done.

TODO:
- [x] Tests
- [x] Documentation
- [ ] Refactor `middleware` to be able to log all status codes and response length
- [ ] Finish `getNscaDateTime` to handle all time zones

I think this is a candidate for a minor release.

Unfortunately, I don't have time to finish this, this year. If you @jfhbrook or someone else have time, I would love to review.

I noticed that the log code path is run two times for each request, which means that the middleware is run two times for each request but I don't understand why we would do that. The second time the `req.method` is `undefined` and `req.url` is encoded.


dotnetCarpenter:
hehe failing in node 4 and 5 after I followed airBnB's eslint - dropping node 4 and 5 support to destructuring assignment or make two small fixes in tests... yaa, remove destructuring - it also sounds so negative.

tap-snapshots/test-cli.js-TAP.test.js
```diff
@@ -0,0 +1,11 @@
+/* IMPORTANT
```
jfhbrook: what is this :o

dotnetCarpenter: http://www.node-tap.org/snapshots/

dotnetCarpenter:
Ref: #208

dotnetCarpenter:
@jfhbrook What do you think about the idea and about refactoring `middleware` - would you be up for doing it? Also do you know why the middleware is run twice for each request?

tap-snapshots/test-cli.js-TAP.test.js
dotnetCarpenter: I added a comment in the test about how to generate the file

dotnetCarpenter:
Looks like codecov is broken somehow.. I'll take a quick look

dotnetCarpenter:
I'm using this branch of ecstatic serve svg and image files to another node program and I never see double logging (re-run of `middleware` for each request), which leads me to believe that this only happens when requesting via a browser. I was using Chromium..
I'm guessing it could be OPTION or prefetch request that we are not handling very well. It's not completely b0rken but pretty bad for logging and probably has other side-effects.

jfhbrook:
I know there's code branches that call the original middleware recursively, those would probably be candidates for ways to cause double-logging.

I'm down to do some middleware refactors in general, given I have the time and the understanding necessary to make useful ones. This might also touch on https://github.com/jfhbrook/node-ecstatic/issues/193 https://github.com/jfhbrook/node-ecstatic/issues/154 https://github.com/jfhbrook/node-ecstatic/issues/88

dotnetCarpenter:
R. I. P.
#259

PR #211, opened 2017-12-16
Fix codecov

dotnetCarpenter:
This remove the node version of codecov and uses the bash version instead.

So the longer any nonsense codecov when doing local `npm test` also I put in an auto job flag for `tap`, so if you got multiple processors, it should speed up tests.

Hurry up and merge this in :) Then I can rebase #209 to master and we see if PR's adds or lower the coverage. But we need at least one coverage report in master before it works.

oh, and I also changed from lcov to clover reports - not sure what the difference is but I found an example with clover and codecov...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,7 @@ node_js:
 install:
   - npm install
   - npm update
+script:
+  - npm test
+after_success:
+    - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


jfhbrook:
Works for me!

I added you as a collaborator, so you should be able to make changes if I die in a ditch. :) I'll try to be around for review and advice though.

Issue #212, opened 2017-12-16
Update mime to a version without a security vulnarability


dotnetCarpenter:
I stumble on this: https://david-dm.org/jfhbrook/node-ecstatic while looking at https://github.com/jfhbrook/node-ecstatic/blob/master/CONTRIBUTING.md#a-few-other-minor-guidelines

That seems pretty bad since it's not a dev dep and we have many hooks to set custom mime types, which someone might set from an entrusted source.

jfhbrook:
oh wow.

questions:

* what version of mime are we on now?
* what are the breaking changes between versions?

jfhbrook:
shit, this means we have to solve https://github.com/jfhbrook/node-ecstatic/issues/66 yeah?

jfhbrook:
I guess we can do the upgrade, major semver ourselves, and kick the can on how we can make mime truly pluggable since I don't think anyone has actually tackled that problem yet

dotnetCarpenter:
Yes there is breaking change. If I update and run `node_modules/.bin/tap test/mime.js` I get a lot of errors.

1. not ok `mime.lookup` is not a function
2. custom definition of mime-type with the mime package
  not ok Attempt to change mapping for "opml" extension from "text/x-opml" to "application/xml". Pass `force=true` to allow this, otherwise remove "opml" from the list of extensions for "application/xml".
3. not ok `mime.load` is not a function

dotnetCarpenter:
@jfhbrook heh I thought I fixed #66 in #143 one and a half year ago ;)
I'm sorry but I really don't have time at the moment. I use ecstatic every day but am behind on project that needs to be wrapped up ASAP. I just needed the log, so I did it. But I hope to have time next year to give ecstatic some :heart:

dotnetCarpenter:
There was an idea about using **mime-magic**. Is that module out?

jfhbrook:
yeah I was thinking about that! I don't remember why I didn't like the idea at the time. Maybe I was worried about cross-platform support? That seems squared with them shipping the windows binary. We already have to do a major change, so that's not an issue either

dotnetCarpenter:
Do we need a coding date in January to refactor mime types, middleware and make ecstatic SPA friendly?

jfhbrook:
ahaha oh man, yeah maybe

dotnetCarpenter:
hehe... I'll be up for a weekend hackathon or something like that if *life* doesn't get in the way.

[...]

PR #218, opened 2018-02-19
Update mime to v2.2.0

mathiasbynens:


jfhbrook:
A bunch of people depend on the mime files capability, though. Do you know of a way to mitigate this?

jfhbrook:
I'm trying to dig up some relevant content and figure out who I have to at

jfhbrook:
The most recent time we punted this https://github.com/jfhbrook/node-ecstatic/issues/212 and I think it links to most of the relevant issues.

I don't actually know how popular the mime files format is, or what it would take to replace it. Maybe we can find some people involved in the mime module we're using to comment on why they removed it?

/cc @dotnetCarpenter @mk-pmb might be relevant to your interests? no obligation to comment etc

jfhbrook:
It should also be noted that a bunch of the tests are failing :point_down:

mathiasbynens:
Yeah, I just wanted to get the ball rolling on this again. I can look into fixing up the PR if there is interest.

> A bunch of people depend on the mime files capability, though. Do you know of a way to mitigate this?

It’s a breaking change for sure. Bumping the major version (per semver) would prevent those depending on this functionality from suffering any breakage.

jfhbrook:
It prevents immediate breakage but also doesn't supply a migration path, which is still pretty unfortunate.

mk-pmb:
I have an unstable branch where my migration strategy is to provide a hook for users to supply their own MIME lookup function, individual for each ecstatic instance. IIRC the old mime package had the problem of keeping a global database for itself, so all modules customizeing "their" MIME would also customize all other modules that use it.

Bonus features I'd like to see soon[TM]:
* Have a per-ecstatic-instance default MIME type option: This string is used if the MIME function returned something false-y.
* In case the MIME function returns something that looks like a then-able, defer serving until it results are in, with reasonable timeout. (late false-y result still means default.)

jfhbrook:
Thenable's a good idea. Otherwise I guess we'd expect the v2 API, include a basic wrapper for the v1 API and drop cli support?

That could be a reasonable compromise.

mk-pmb:
> we'd expect the v2 API

In which way "expect"? We'd use the MIME v2 API for the default lookup of course.

If users want to customize that lookup in any way, they can provide a function that looks up a MIME type by filename. If we're generous we can supply additional context about the FTP session as a 2nd argument. If they want to use the MIME module under the hood, its API is their problem.

mk-pmb:
Sorry for the mixum. I meant the ecstatic instance, there's no FTP here. :-)

mk-pmb:
Even better compromise: If the user MIME function yields a false-y result, that means ecstatic shall figure out on its own, using its default strategy (currently: mime v2).
This way, a user function for some few specials could be as easy as

```javascript
function (filename) {
  return ({
    txt: 'text/comma-separated-values',
    ico: 'image/bitmap',
  })[filename.match(/\.(\w+)$|/)[1]];
}
```

jfhbrook:
oh I gotcha, we'd do `mineLookupFunction: ((fname) => mime2.getType(fname))` or some such as a default and let people set that themselves. I like it!

I think as far as accepting thenables, I think that can be done in a separate PR.

@mathiasbynens if this proposal makes sense to you, and I'm not missing something, and you want to implement this rq (it shouldn't be too bad, the options parsing is a bit involved but easy to copy), I'd merge it.

mk-pmb:
As the default for `mineLookupFunction` I'd use either `null` (means ecstatic shall not even try to use the custom func) or a function that always returns a false-y value (e.g. `(() => {})`). If we're going to expose our default lookup mechanism, we should so so via an explicit API, not a config default.

jfhbrook:
without a default it would break, no? There's no behavior difference here.

mk-pmb:
Yes, behavior is the same, it's just ninja protection. Avoiding to expose a useful MIME handler is just help people not be overly "clever" by depending on an accidential API feature.

BigBlueHat:
Is the key blocker here the regression of the ability to load a custom `.types` file (i.e. the `.load()` system being removed from `mime` 2.x)?

The [`mime-db`](https://github.com/jshttp/mime-db) project (from where `mime` gets its JSON type files) contains some code for parsing (via regex) and writing the JSON format that `mime-db` and `mime` use:
https://github.com/jshttp/mime-db/blob/master/scripts/fetch-apache.js

There are two conceivable approaches (so far :wink:):
1. add this sort of code and continue to handle `.types` file references (converting the them into the JSON `mime` needs at runtime)
2. ship a script which converts `.types` to these `.json` files and then passing those `.json` file references in via the command line

My preference would be for option 1 as the command line "signature" and expectations would be the same, but I can understand the motivations behind option 2.

Thoughts?

jfhbrook:
There's also

3. Lose the built in cli capability of parsing .types files, but add the option of supplying a custom mime lookup function which may or may not leverage mime@1

which was suggested by mk-pmb, who iirc is the most vocal person around this sort of functionality, so it works for me.

guybedford:
This would be good to support WASM, as it still doesn't serve in Firefox with this server as far as I can tell.

jfhbrook:
Closing in favor of #240

Issue #214, opened 2018-01-06
Add way to hide file permissions


benevbright:
Thanks for a wonderful module.

I want to hide files' permission.
Is there any solution for this?
  
I can't find it from the provided options.

jfhbrook:
It's not a currently exposed option, but I would definitely take this as a pull request!

benevbright:
Thanks for your kind response :)

mmmm1998:
I think, the issue maybe closed, isn't it?

jfhbrook:
yes, just forgot.

benevbright:
Thank @mmmm1998 . I came by now because I thought I have time to deal with. 
But it's already done! I appreciate it.

benevbright:
I made an advanced version of @mmmm1998 's PR.
I appreciate to @mmmm1998 .

https://github.com/jfhbrook/node-ecstatic/pull/219

PR #219, opened 2018-02-23
Providing options that developer has a choice what to show to the user.

benevbright:
It started from #214 .

and @mmmm1998 gave a hand.

I made a change little more. 
It provides options that developer has a choice what to show and those order to the user.

> selectDisplay:"icon|permissions|filesize|filename"

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ const opts = {
   showDir: true,
   showDotfiles: true,
   humanReadable: true,
-  selectDisplay:"ipsn",
+  selectDisplay:"icon|permissions|filesize|filename",
   si: false,
   cache: 'max-age=3600',
   cors: false,
--- a/README.md
+++ b/README.md
@@ -159,6 +159,11 @@ Exclude dotfiles from directory listings with `opts.showDotfiles === false`. Def
 If showDir is enabled, add human-readable file sizes. Defaults to **true**.
 Aliases are `humanreadable` and `human-readable`.
 
+### `opts.selectDisplay`
+### `--selectDisplay` {WhatToDisplay}
+
+If you set selectDisplay with **'icon|permissions|filesize|filename'**, you can choose what to display and how order them.
+
 ### `opts.headers`
 ### `--H {HeaderA: valA} [--H {HeaderB: valB}]`
 
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ const opts = {
   showDir: true,
   showDotfiles: true,
   humanReadable: true,
-  hidePermissions: false,
+  selectDisplay:"ipsn",
   si: false,
   cache: 'max-age=3600',
   cors: false,
@@ -158,11 +158,6 @@ Exclude dotfiles from directory listings with `opts.showDotfiles === false`. Def
 
 If showDir is enabled, add human-readable file sizes. Defaults to **true**.
 Aliases are `humanreadable` and `human-readable`.
-
-### `opts.hidePermissions`
-### `--hide-permissions`
-If hidePermissions is enabled, file permissions will not be displayed. Defaults to **false**.
-Aliases are `hidepermissions` and `hide-permissions`.
 
 ### `opts.headers`
 ### `--H {HeaderA: valA} [--H {HeaderB: valB}]`
--- a/lib/ecstatic/defaults.json
+++ b/lib/ecstatic/defaults.json
@@ -3,7 +3,7 @@
   "showDir": true,
   "showDotfiles": true,
   "humanReadable": true,
-  "hidePermissions": false,
+  "selectDisplay":"icon|permissions|filesize|filename",
   "si": false,
   "cache": "max-age=3600",
   "cors": false,
--- a/lib/ecstatic/show-dir/index.js
+++ b/lib/ecstatic/show-dir/index.js
@@ -110,8 +110,8 @@ module.exports = (opts) => {
             // TODO: use stylessheets?
             html += '<tr>';
             const selectDisplayOptions = selectDisplay.split('|');
-            for(let i = 0 ; i < selectDisplayOptions.length ; i++){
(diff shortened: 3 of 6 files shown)


benevbright:
I want to hear what you guys think.

mmmm1998:
Well, what about CLI version? And I think it would be nice to fix travis's error. Add information about options in README.md is usefull too, I think.

benevbright:
It’s working with CLI version correctly.
Yes I will do that anyway. Thanks.

benevbright:
Everything is done. 
I would really appreciate if someone can correct my bad English in Readme.

lib/ecstatic/show-dir/index.js
```diff
@@ -108,15 +108,27 @@ module.exports = (opts) => {
             const iconClass = `icon-${isDir ? '_blank' : classForNonDir}`;
 
             // TODO: use stylessheets?
-            html += `${'<tr>' +
-              '<td><i class="icon '}${iconClass}"></i></td>`;
-            if (!hidePermissions) {
-              html += `<td class="perms"><code>(${permsToString(file[1])})</code></td>`;
+            html += '<tr>';
+            const selectDisplayOptions = selectDisplay.split('|');
+            for (let i = 0; i < selectDisplayOptions.length; i += 1) {
```
jfhbrook: This should be `selectDisplayOptions.forEach(function(displayOption) {`

lib/ecstatic/defaults.json
```diff
@@ -3,7 +3,7 @@
   "showDir": true,
   "showDotfiles": true,
   "humanReadable": true,
-  "hidePermissions": false,
+  "selectDisplay":"icon|permissions|filesize|filename",
```
jfhbrook: I think this API should take an object:

```js
{ icon: true, permissions: true, filesize: true, filename: true }
```

something like this. On the CLI, I could be convinced to take on this API, though we have an existing mime types API that takes JSON on the CLI, so that's worth considering...

lib/ecstatic/show-dir/index.js
```diff
@@ -108,15 +108,27 @@ module.exports = (opts) => {
...
+              switch (selectDisplayOptions[i].trim()) {
+                case 'icon':
+                  html += `${'<td><i class="icon '}${iconClass}"></i></td>`;
```
jfhbrook: This approach means we can use display options to change the *order* in which elements are displayed. I think we can fix this by iterating over a hard-coded list order. `['icon', 'permissions', 'filesize', 'filename'].forEach(function(option) {`

README.md
```diff
@@ -159,10 +159,10 @@ Exclude dotfiles from directory listings with `opts.showDotfiles === false`. Def
 If showDir is enabled, add human-readable file sizes. Defaults to **true**.
 Aliases are `humanreadable` and `human-readable`.
 
-### `opts.hidePermissions`
-### `--hide-permissions`
-If hidePermissions is enabled, file permissions will not be displayed. Defaults to **false**.
-Aliases are `hidepermissions` and `hide-permissions`.
+### `opts.selectDisplay`
```
jfhbrook: Kind of funky naming. It sounds like English isn't your first language so that's fine. I'll try to brainstorm something I like. :)

jfhbrook:
Having thought about this, I think we should refactor to go better with #215 by leveraging the existing flag for permissions, and adding `hideIcons` and `hideFilesize` as separate options.

benevbright:
I was going to answer to your review. By the way, thanks for your review. I appreciate it. 
Your review makes sense to me. 
Yeap, I suppose you would want to keep simple approach. I agree that too.
All right then. :)

benevbright:
And I guess we better consider refactoring later when there is someone who thinks hiding icons and filesize are necessary.
Even though I made a PR, I was not sure if it's necessary.

README.md
benevbright: Yeap, you are right :)

lib/ecstatic/show-dir/index.js
benevbright: Yes. I meant it. 
But how to get an order since it should be an object? I believe the object does not guarantee its order.

jfhbrook: correct--afaik while in practice objects are sorted that the language spec doesn't include that property. So!

There are multiple ways to skin this, but maybe something like

```js
[
  ['hideIcons', `${'<td><i class="icon '}${iconClass}"></i></td>`],
  ['hidePermissions', `<td class="perms"><code>(${permsToString(file[1])})</code></td>`]
].forEach(function(pair) {
  if (opts[pair[0]]) {
    html += pair[1]
  }
});
```

A little gross maybe, and there's more than one way to skin this one for sure, but this is a strategy that might work.

jfhbrook: This reminds me of something though--I'll mark it in a separate comment.

jfhbrook:
Yeah that's reasonable. :) Thanks for the suggestion!

lib/ecstatic/show-dir/index.js
mmmm1998: Well, what about templated html-page? I mean, in templated html we have strong order of html-elements. So, templated html-page could have something like this:
```handlebars
<table>
  {{#each files}}
    <tr>
      {{#if icon}}
        <td>
          <i class="icon {{iconType}}"></i>
        </td>
      {{/if}}
      {{#if permissions}}
        <td class="perms">
          <code>{{permissions}}</code>
        </td>
      {{/if}}
      {{#if filesize}}
        <td class="file-size">
            <code>{{size}}</code>
        </td>
      {{/if}}
      {{#if filename}}
        <td class="display-name">
          <a href={{href}}>{{name}}</a>
        </td>
      {{/if}}
    </tr>
  {{/each}}
</table>
```
But templated html page might be overkill for this task, so i don't know.

PR #221, opened 2018-04-27
Fix 416 if there are spaces in range header

victordidenko:
According to [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html):

> 4.2 Message Headers
> ...
> The field-content does not include any leading or trailing LWS: linear white space occurring before the first non-whitespace character of the field-value or after the last non-whitespace character of the field-value. Such leading or trailing LWS MAY be removed without changing the semantics of the field value.

But if header `Range` doesn't contain \<range-end\>, but contains space after `-` → then `parseInt(' ', 10)` returns NaN and ecstatic responses with 416 Range Not Satisfiable.

Real client request and response example:
screen shot 2018-04-27 at 22 05 54
screen shot 2018-04-27 at 22 06 20

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -262,7 +262,7 @@ module.exports = function createMiddleware(_dir, _options) {
 
       if (range) {
         const total = stat.size;
-        const parts = range.replace(/bytes=/, '').split('-');
+        const parts = range.trim().replace(/bytes=/, '').split('-');
         const partialstart = parts[0];
         const partialend = parts[1];
         const start = parseInt(partialstart, 10);


jfhbrook:
Thanks!

victordidenko:
No problem :)

Also `Range` header has [more complex format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range), then just `bytes=<start>-<end>`, but I am not ready yet to make PR to support it fully :) This PR is fixing error I met in practise, as you can see from screenshots from wireshark.

jfhbrook:
Yeah I feel you. I'll definitely accept it if you wanna flesh out range requests some more!!

Issue #44, opened 2012-07-29
include charset in Content-Type header


paulwalker:

markandrus:
I know this is an old thread, but it would be very useful to be able to override the charset. mime's charset lookup function is basically "return 'utf-8' if it has 'text' appears in the MIME type", which is incorrect for many applications.

jfhbrook:
It's late my time and idk when I'll get a chance to dig into the details of your request, but this sounds like something I would accept as a PR if you want to take a crack at it.

dotnetCarpenter:
@markandrus Can't you already do that with [opts.mimeTypes](https://github.com/jfhbrook/node-ecstatic#optsmimetypes)?

dotnetCarpenter:
Looks like you can not, due to https://github.com/broofa/node-mime/blob/master/mime.js#L104 and https://github.com/jfhbrook/node-ecstatic/blob/master/lib/ecstatic.js#L201.

or.... what does `ecstatic --mime-types "{ mimeType: { 'what/ever;charset=us-ascii': ['php']  }` do?

PR #223, opened 2018-06-19
Remove charset=utf-8 default

### LINKED ISSUE #220: Wrong behavior of `application/wasm` mime type
Current webassebmly spec says
> Note: extra parameters are not allowed, including the empty `application/wasm;`.
https://webassembly.github.io/spec/web-api/index.html#streaming-modules

For now ecstatic adds charset in `Content-Type` for all filetypes, so mimetype for wasm files will be `application/wasm; charset=utf-8`. Can we somehow disable this for files with `application/wasm` mimetype (or for binary files at all?)?

I'll be happy to create PR if you tell me right way to figure out exclusions.

BigBlueHat:
Fixes #220 

This also fixes image loading issues when using `fetch()` + `createImageBitmap()` in Firefox:
https://github.com/indexzero/http-server/issues/296#issuecomment-338352458

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -240,7 +240,7 @@ module.exports = function createMiddleware(_dir, _options) {
       let stream = null;
 
       if (contentType) {
-        charSet = mime.charsets.lookup(contentType, 'utf-8');
+        charSet = mime.charsets.lookup(contentType);
         if (charSet) {
           contentType += `; charset=${charSet}`;
         }
--- a/test/cli.js
+++ b/test/cli.js
@@ -115,7 +115,7 @@ test('setting mimeTypes via cli - directly', (t) => {
     t.pass('ecstatic should be started');
     checkServerIsRunning(`${defaultUrl}:${port}/custom_mime_type.opml`, t, (err, res) => {
       t.error(err);
-      t.equal(res.headers['content-type'], 'application/x-my-type; charset=utf-8');
+      t.equal(res.headers['content-type'], 'application/x-my-type');
     });
   });
 });
--- a/test/custom-content-type-file-secret.js
+++ b/test/custom-content-type-file-secret.js
@@ -24,7 +24,7 @@ test('custom contentType via .types file', (t) => {
     request.get(`http://localhost:${port}/custom_mime_type.opml`, (err, res) => {
       t.ifError(err);
       t.equal(res.statusCode, 200, 'custom_mime_type.opml should be found');
-      t.equal(res.headers['content-type'], 'application/secret; charset=utf-8');
+      t.equal(res.headers['content-type'], 'application/secret');
       server.close(() => { t.end(); });
     });
   });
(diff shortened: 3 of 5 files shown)


jfhbrook:
Thanks for trying to make ecstatic better! I agree that it's alarming that images break because of this. A few minor concerns:

1. This would be a major version bump as-is, which is fine in and of itself but I dunno that I'd consider this a patchfix of the main line. If this was feature flagged somehow (or made configurable s.t. you can set your own charsets) I could justify it being a minor, and then flip the defaults for a major later.
2. You mentioned being interested in upgrading the mime package--would that fix this "better"?

BigBlueHat:
> made configurable s.t. you can set your own charsets

I wouldn't want to encourage a blanket `charset` setting option as that puts us back right where we are (passing `charset` on all the things).

Maybe an `--auto-charset` option would do the trick? It could become the default in a future major version and the defaulting to `utf-8` would be gone.

> You mentioned being interested in upgrading the mime package--would that fix this "better"?

See https://github.com/jfhbrook/node-ecstatic/issues/220#issuecomment-398443210 :)

athei:
I also stumbled on this because it prevents Firefox from using wasm streaming instantiation. Why is this a major version bump? We are not breaking the API. It is a bug fix isn't it?

jfhbrook:
One person's bugfix is another's expected behavior 🤷‍♂️

athei:
Okay :D 

So the progress on this is stalling because you need to evaluate if you skip this patch alltogether and just apply #218 instead?

jfhbrook:
I'm not merging #218, at least not as-is, it's open for discussion and so that someone can take the opportunity to fix the issues outlined in that PR.

FWIW I /believe/ this ticket is still an issue even with something along the lines of #218 merged.

jfhbrook:
Dredged up https://github.com/jfhbrook/node-ecstatic/issues/44 which was the original issue--I think there are good reasons to specify the charset, hence why this exists in the first place.

athei:
Yes and this patch delegates that charset decision to the mime library like it should. The charset should still be added to types that need it after that patch.

jfhbrook:
A test that confirms that would be nice!

jfhbrook:
Anyway, just to reiterate what I'm looking for around this and #44:

* Some answer to cases where setting charset would be appropriate--either delegating to an autodetect library, or an optional way to override, with tests
* Some answer to passing in custom types--either a separate library for parsing types files with new mime, or the ability to pass in a custom mimetypes resolver in javascript

jfhbrook:
I get that you're frustrated, btw, with this long-standing issue that prevents a lot of cool things from working right. The problem is that I don't want to merge casual code, but don't have the time to implement the right thing myself. All I can say is, be the change you wish to see in the world.

jfhbrook:
I rolled these changes into #240, which addresses the concerns I had by exposing a way to set a custom content-type lookup function. Thanks!

PR #225, opened 2018-07-09
Fix bad parsing on CORS option

s25g5d4:
Hi there, it's me again. As I metioned last time I was poking around the headers returned by node-ecstatic, I noticed that the CORS-related headers were set without any option provided by me. And even if I turn the option off explicitly, they still exist. It is not hard to find the problem, but I'm suprised that the problem exists for such a long time.

I also add a test, and hopefully that helps. I don't have much experience on writing tests.

Thanks for reviewing.

### DIFF
--- a/lib/ecstatic/opts.js
+++ b/lib/ecstatic/opts.js
@@ -114,7 +114,7 @@ module.exports = (opts) => {
     });
 
     aliases.cors.forEach((k) => {
-      if (isDeclared(k) && k) {
+      if (isDeclared(k) && opts[k]) {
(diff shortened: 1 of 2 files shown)


jfhbrook:
Oh neat! My guess is that only a handful of people were using this feature, and I suspect most of them are using the --cors flag.

It's bedtime for me but I can probably look at this tomorrow. Thanks!

jfhbrook:
Closing in favor of #241, which is just this with some minor tweaks and a changelog update.

Thanks a bunch! Sorry this took so long to pull in.

PR #227, opened 2018-08-28
Trim whitespace from request encoding headers

thornjad:
This patch simply trims whitespace when checking the `accept-encoding` header. This header is split on commas, but this leaves a leading space on every element except the first, at least in Firefox and Chromium. By coincidence, this has been working because the first encoding in the list matches right away, but this behavior is not guaranteed.

Another way this could be done would be to split on `', '` (comma and space), and this would work in Firefox and Chromium for sure, but I don't believe this is guaranteed either. Using `.trim()` is safer since `'foo' === 'foo'.trim()`.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -38,7 +38,7 @@ function shouldCompress(req) {
   return headers && headers['accept-encoding'] &&
     headers['accept-encoding']
       .split(',')
-      .some(el => ['*', 'compress', 'gzip', 'deflate'].indexOf(el) !== -1)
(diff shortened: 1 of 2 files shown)


jfhbrook:
Gotcha, good find!

Any interest in adding a test? :)

thornjad:
Yeah, a test would be a good idea, I should have started with that! Coming right up, chef! :)

jfhbrook:
Sick! Lemme know if the test suite is confusing at all

thornjad:
Two tests added! Let me know if you think it needs any deeper testing.

jfhbrook:
lgtm! Thanks.

jfhbrook:
Published 3.2.2

PR #229, opened 2018-08-31
Set Brotli compression header

filipesilva:
This PR is similar to https://github.com/jfhbrook/node-ecstatic/pull/228 but instead of compressing using Brotli, it just adds the header if a Brotli compressed file is found.

/cc @alexeagle


filipesilva:
This isn't needed anymore since https://github.com/jfhbrook/node-ecstatic/pull/228 went in 👍

PR #237, opened 2018-10-30
eliminate baseDir and support mounting, close #235

### LINKED ISSUE #235: baseDir does not work correctly when mounted non-root, and is not necessary
Previously I posted issue #233 , but now I think that is not about absolute path/relative path, but the way this middleware works.
The goal may be achieved by using the middleware like this:
`app.use('/', ecstatic({.., baseDir: "a"}));`,
However, if I use it like this:
`app.use('/a', ecstatic({.., baseDir: "a"}));	//no matter whether baseDir is set`,
it does not work correctly as I addressed in #233 , but using a middleware like that is quite common in express apps.
Think a little further, due to express's mounting feature, as stated in [req.originalUrl](http://expressjs.com/en/4x/api.html#req.originalUrl), `baseDir` might be unnecessary at the 1st place. The user can just mount the ecstatic at the path they want, and let express strip the base for you:
`app.use('/a', ecstatic({..}));`
This works better, because this makes the middleware more portable, in the sense that the middleware does not need to know which path is it working on. The middleware can just assume the `req.url` contains barely the file path, which simplifies things a lot.
How do you think? @jfhbrook

panlina:
As addressed in #235 , `baseDir` is eliminated and mounting is supported to achieve the same purpose.

Instead of
`app.use(ecstatic({.., baseDir: "a"}));`,
we now do:
`app.use('/a', ecstatic({..}));`.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -67,3 +67,4 @@ Listed in no particular order:
 * Victor Didenko @victordidenko
 * Zong Jhe Wu @s25g5d4
 * Jade Michael Thornton @thornjad <jade@jmthorton.net>
+* Jack Lu @panlina <jacklu@jacklu.me>
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ In node, pass ecstatic an options hash, and it will return your middleware!
 ```js
 const opts = {
   root: path.join(__dirname, 'public'),
-  baseDir: '/',
   autoIndex: true,
   showDir: true,
   showDotfiles: true,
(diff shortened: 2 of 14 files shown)


panlina:
@jfhbrook The codecov check fails. Does that mean I need add tests? I'm not familiar with coverage tools and how they count. You can review my changes and tell me if there's anything I need to do.

imcuttle:
@panlina I guess you do not need the complex processing, 

Just wrap ecstatic simply by https://github.com/jfhbrook/node-ecstatic/issues/235#issuecomment-433590144

panlina:
@imcuttle It's not "complex processing". It's simplification. 😄 . I saw your solution. It may work but that's a temporary solution as you said.

jfhbrook:
Forgive my thread necromancy!

My concern with this PR as it stands is that a lot of people use this module outside the context of ecstatic, where this flag is actually pretty useful.

Unfortunately I don't use express very often and certainly don't use this module with express! so I don't actually understand the desired behavior with express that well or why basedir breaks it. Is there a way to get this behavior *without* removing the baseDir option?

jfhbrook:
Something that might be helpful for me is breaking tests for the behavior in express - this would give me a spec to work against.

panlina:
Good to see you back after months!

Actually I changed my mind after that. It's better to keep it not relying on a hosting framework as since it's born.

Thank you for taking care of this PR. Now you can close it.

PR #246, opened 2019-04-11
fix: leak on res aborted on non-range requests

ronag:
Currently a non-range request which is aborted will leak a file descriptor. This PR fixes that.


jfhbrook:
Thanks!

PR #247, opened 2019-04-13
fix: handle some edge cases with errors

ronag:
You can't write headers if headers are already sent... which is the case if e.g. the read stream fails (emit error) AFTER initial byte.

I'm not sure checking `writable` is enough and also if headers are sent we still need to destroy the response.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -296,7 +296,11 @@ module.exports = function createMiddleware(_dir, _options) {
 
         fstream = fs.createReadStream(file, { start, end });
         fstream.on('error', (err) => {
-          status['500'](res, next, { error: err });
+          if (handleError) {
+            status['500'](res, next, { error: err });
+          } else {
+            next(err);
+          }
         });
         res.on('close', () => {
           fstream.destroy();
@@ -342,7 +346,11 @@ module.exports = function createMiddleware(_dir, _options) {
 
       stream.pipe(res);
       stream.on('error', (err) => {
-        status['500'](res, next, { error: err });
+        if (handleError) {
+          status['500'](res, next, { error: err });
+        } else {
+          next(err);
+        }
       });
     }
 
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -296,11 +296,7 @@ module.exports = function createMiddleware(_dir, _options) {
 
         fstream = fs.createReadStream(file, { start, end });
         fstream.on('error', (err) => {
-          if (handleError) {
-            status['500'](res, next, { error: err });
-          } else {
-            next(err);
-          }
+          status['500'](res, next, { error: err, handleError });
         });
         res.on('close', () => {
           fstream.destroy();
@@ -346,11 +342,7 @@ module.exports = function createMiddleware(_dir, _options) {
 
       stream.pipe(res);
       stream.on('error', (err) => {
-        if (handleError) {
-          status['500'](res, next, { error: err });
-        } else {
-          next(err);
-        }
+        status['500'](res, next, { error: err, handleError });
       });
     }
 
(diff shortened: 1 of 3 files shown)


lib/ecstatic/status-handlers.js
```diff
@@ -2,58 +2,58 @@
 
 const he = require('he');
 
+function wrap(statusCode, fn) {
+  return (res, next, opts) => {
+    if (typeof next === 'function') {
```
jfhbrook: This is inconsistent with the current behavior, right?

By my research, it looks like current behavior is to only check handleError in 404 cases and that it's not wired up literally anywhere else, and that prior to this change "next" was ignored, even though it was being passed in. 🤔 Not sure what the intent was there.

So there are a couple of things that would be good to see, if you're game:

- [ ] A documentation update to make this behavior more clear - shouldn't be a big deal, just describing that handleError: false will fall-through for whatever set of status codes
- [ ] Some tests that check that next is called properly depending on the state of handleError. I realize that the tests are kind of a shitshow. If this is a relatively low lift 🤞 great! If not, maybe not a big deal.

ronag: Rather than making things complicated... I think I've fixed it so that it is consistent with the current behavior.

ronag: I have to admit I don't understand the current behavior... but let's fix the current behavior first and then consider changing it...

jfhbrook: Honestly, I don't either. This is a project I wrote a long time ago and I haven't held the whole thing in my head for a while. Honestly there are a lot of things I would do different in a rewrite, but the test harness is awkward enough that I think a refactor that size would be challenging.

Anyway, ignoring next(), or removing its use from the call sites is fine by me, though one of us should flag this as an issue.

jfhbrook: Written up here: https://github.com/jfhbrook/node-ecstatic/issues/251 feel free to add notes if you have any.

ronag:
Ok, I tried cleaning this up a bit and make it behave as "expected".

ronag:
Please merge https://github.com/jfhbrook/node-ecstatic/pull/249 first as I believe it will cause some conflicts with this PR that needs to be resolved first.

lib/ecstatic/status-handlers.js
```diff
@@ -2,58 +2,56 @@
...
+  res.end('ACCESS DENIED');
+});
 
 // disallowed method
-exports['405'] = (res, next, opts) => {
-  res.statusCode = 405;
-  if (typeof next === 'function') {
-    next();
-  } else {
-    res.setHeader('allow', (opts && opts.allow) || 'GET, HEAD');
```
ronag: Note that I removed `opts.allow`... since it wasn't used anywhere

lib/ecstatic.js
```diff
@@ -353,7 +353,7 @@ module.exports = function createMiddleware(_dir, _options) {
           if (req.statusCode === 404) {
             // This means we're already trying ./404.html and can not find it.
             // So send plain text response with 404 status code
-            status[404](res, next);
+            status[404](res);
```
ronag: @jfhbrook: This might need an implicit next instead of `status[404]`?

e.g. I would suggest for *404 only*

```js
if (next) {
  next()
} else {
   status[404](res)
}
```

ronag:
The problem as I see it is that this library tries to act both like a middleware and an actual server so it is very difficult to figure out how to handle cases such as 404 (should call next if middleware, but end if server) etc...

ronag:
The scope of this becomes too large... I've changed the PR to minimal changes and I believe it should be safe to merge without breaking anyone.

ronag:
Actually I think I figured it out pretty well with minimal changes.

if `next` is thruthy, act like a middleware.
if `next` is falsy, act like a server.

Consistent behaviour only required a few changes which I believe to be rather appropriate.

jfhbrook:
Merged manually and released as 4.1.1. Thanks!

PR #249, opened 2019-04-13
fix: on-finished

ronag:
Use on-finished. `close` is not always emitted in every scenario in some "older" versions of node. I don't remember the exact specifics.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -242,6 +242,10 @@ module.exports = function createMiddleware(_dir, _options) {
 
 
     function serve(stat) {
+      if (onFinished.isFinished(res)) {
+        return;
+      }
+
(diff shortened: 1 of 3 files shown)


ronag:
An alternative which is also (as compatible as I can think of is)

```js
function onFinished(req, res, cb) {
      req
        .on('close', cb)
        .on('aborted', cb)
        .on('error', cb)
      res
        .on('close', cb)
        .on('finish', cb)
        .on('error', cb)
}
```

jfhbrook:
I have no problem bringing in a new dependency, this project has not very many as it is.

If I were to merge this, which PRs would this be an alternative to? It looks like this handles a case that #250 doesn't.

ronag:
This is ready for merge.

jfhbrook:
Released as 4.0.2. Thanks!

PR #250, opened 2019-04-13
fix: response can be aborted during e.g. fs.stat and hasGzipId12

ronag:
Request/Response can be aborted while doing async stuff... handle it.

Would prefer to use `isFinished` but I've proposed using the `on-finished` package in another PR so I'll leave the "modern" node way here.

### DIFF
--- a/lib/ecstatic.js
+++ b/lib/ecstatic.js
@@ -239,8 +239,12 @@ module.exports = function createMiddleware(_dir, _options) {
       return;
     }
 
-
     function serve(stat) {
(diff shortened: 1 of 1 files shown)


ronag:
I honestly have no idea how to fix the code coverage on this...

jfhbrook:
I'm honestly not sure how to test this either, these kinda time-dependent things are tough! I say don't sweat the codecov report, I'm not 100% taking it to heart.

ronag:
I implemented this in the on-finished PR with onFinished.isFinished.

PR #261, opened 2019-06-05
Fix redirections on Windows

Xmader:
Thank you for fixing Backport Open Redirect vulnerability, but it breaks on Windows machines. 

`path.normalize("/path/to/file")` returns `\path\to\file` on Windows, and it doesn't work in browsers.


Xmader:
https://github.com/node-ecstatic/node-ecstatic

```
npm i @node-ecstatic/ecstatic
```

PR #266, opened 2020-04-01
v4.1.3: [security] Fix crash on redirect with formfeed in URL (CVE-2019-10775)

mk-pmb:
Sorry for disturbing the peace of the dead, but I'd really appreciate you making an exception so I don't need to start a maintenance fork. (If you wish I do I might consider it though.)

> This fixes CVE-2019-10775. I'll postpone regression tests because
> the request library cannot send those requests (crashing the client
> instead of ecstatic), so I'll need to find another module that can.


mk-pmb:
Thanks!

jfhbrook:
out as 4.1.4 due to having to fuck with github actions but it should be out

mk-pmb:
… and only now it comes to my mind that the RegExp should probably have matched U+0020 space as well. However the impact is probably limited to confusing readers of that code. I'll probably fix that once I get to write the tests for it.

